PR #243, opened 2024-05-22
Update run.py

dorbaker:
Let's remove these lines so that indexes can easily be updated with more documents. Rely on caching to prevent duplication of work.


JoedNgangmeni:
Does this have something to do with the --resume feature? How do we use it? I had an error while indexing and it did not generate_text_embeddings. I wonder if i can run the program from where i left off

PR #332, opened 2024-07-03
Update indexing cli

jgbradley1:
## Description
When using the indexing CLI, the `--root` directory should have a default value of the current directory. Also a new parameter should be added that allows users to easily overlay default configuration values on top of a custom config file. This will allow users to easily specify custom configuration settings in a config file but assume default configuration behavior for the rest of the configuration file.

## Proposed Changes

* set the root to have a default value of "."
* add a new `--overlay-defaults` flag that can be used to 

## Checklist

### DIFF
--- a/.semversioner/next-release/patch-20240703152422358587.json
+++ b/.semversioner/next-release/patch-20240703152422358587.json
@@ -1,0 +1,4 @@
+{
+  "type": "patch",
+  "description": "Add cli flag to overlay default values onto a provided config."
+}
--- a/graphrag/index/__main__.py
+++ b/graphrag/index/__main__.py
@@ -28,7 +28,7 @@ if __name__ == "__main__":
     )
     parser.add_argument(
         "--root",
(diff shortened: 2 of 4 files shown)


AlonsoGuevara (approved):

AlonsoGuevara (approved):

zanderjiang:
Hi about this, I have a config.json file in my data project root. I do actually have the full configuration in the json file, with additional changes. 

When I runt he cli command "python -m graphrag.index --root ./grag," the log indicates it's "graphrag.index.cli INFO using default configuration:" still using the default configuration. 

I'm running a different model and need to adjust the top_p/temperature values in order for it to work.

I've tried setting GRAPHRAG_LLM_TEMPERATURE and GRAPHRAG_LLM_TOP_P in the .env file but it also doesn't seem to load, hence why I tried a custom config.

zanderjiang:
Hi when will this be available?

Issue #357, opened 2024-07-04
Error using vllm OpenAI server


Shamepoo:
INFO:     172.16.80.35:48532 - "POST /v1/chat/completions HTTP/1.1" 500 Internal Server Error
ERROR:    Exception in ASGI application
Traceback (most recent call last):
  File "/home/bigdata/anaconda3/envs/vllm/lib/python3.11/site-packages/uvicorn/protocols/http/httptools_impl.py", line 411, in run_asgi
    result = await app(  # type: ignore[func-returns-value]
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/bigdata/anaconda3/envs/vllm/lib/python3.11/site-packages/uvicorn/middleware/proxy_headers.py", line 69, in __call__
    return await self.app(scope, receive, send)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/bigdata/anaconda3/envs/vllm/lib/python3.11/site-packages/fastapi/applications.py", line 1054, in __call__
    await super().__call__(scope, receive, send)
  File "/home/bigdata/anaconda3/envs/vllm/lib/python3.11/site-packages/starlette/applications.py", line 123, in __call__
    await self.middleware_stack(scope, receive, send)
  File "/home/bigdata/anaconda3/envs/vllm/lib/python3.11/site-packages/starlette/middleware/errors.py", line 186, in __call__
    raise exc
  File "/home/bigdata/anaconda3/envs/vllm/lib/python3.11/site-packages/starlette/middleware/errors.py", line 164, in __call__
    await self.app(scope, receive, _send)
  File "/home/bigdata/anaconda3/envs/vllm/lib/python3.11/site-packages/starlette/middleware/cors.py", line 85, in __call__
    await self.app(scope, receive, send)
  File "/home/bigdata/anaconda3/envs/vllm/lib/python3.11/site-packages/starlette/middleware/exceptions.py", line 65, in __call__
    await wrap_app_handling_exceptions(self.app, conn)(scope, receive, send)
  File "/home/bigdata/anaconda3/envs/vllm/lib/python3.11/site-packages/starlette/_exception_handler.py", line 64, in wrapped_app
    raise exc
  File "/home/bigdata/anaconda3/envs/vllm/lib/python3.11/site-packages/starlette/_exception_handler.py", line 53, in wrapped_app
    await app(scope, receive, sender)
[...]

PR #390, opened 2024-07-05
Add N parameter support

AlonsoGuevara:
## Description

- Add N parameter support
- Add new parameters to init config
- update docs

## Related Issues

#357 

## Checklist


docsite/posts/config/env_vars.md
```diff
@@ -59,11 +59,14 @@ These settings control the text generation model used by the pipeline. Any setti
 | `GRAPHRAG_LLM_THREAD_COUNT`                       |                          | The number of threads to use for LLM parallelization.                          | `int`   | 50                    |
 | `GRAPHRAG_LLM_THREAD_STAGGER`                     |                          | The time to wait (in seconds) between starting each thread.                    | `float` | 0.3                   |
 | `GRAPHRAG_LLM_CONCURRENT_REQUESTS`                |                          | The number of concurrent requests to allow for the embedding client.           | `int`   | 25                    |
-| `GRAPHRAG_LLM_TPM`                                |                          | The number of tokens per minute to allow for the LLM client. 0 = Bypass        | `int`   | 0                     |
-| `GRAPHRAG_LLM_RPM`                                |                          | The number of requests per minute to allow for the LLM client. 0 = Bypass      | `int`   | 0                     |
+| `GRAPHRAG_LLM_TOKENS_PER_MINUTE`                  |                          | The number of tokens per minute to allow for the LLM client. 0 = Bypass        | `int`   | 0                     |
```
natoverse, line 62: How do we want to characterize param changes like this? Seems breaking (e.g., major bump) unless we build in a fallback to the old one for back-compat

darthtrevino (approved):
Can you verify that the `n` argument updates the LLM cache key? If it doesn't this could lead to bad cache responses.

docsite/posts/config/env_vars.md
natoverse: @darthtrevino ?

docsite/posts/config/json_yaml.md
```diff
@@ -67,6 +67,9 @@ This is the base LLM configuration section. Other steps may override this config
 - `max_retry_wait` **float** - The maximum backoff time.
 - `sleep_on_rate_limit_recommendation` **bool** - Whether to adhere to sleep recommendations (Azure).
 - `concurrent_requests` **int** The number of open requests to allow at once.
+- `temperature` **float** - The temperature to use.
+- `top_p` **float** - The top-p value to use.
+- `n` **int** - The number of completions to generate.
```
natoverse, line 72: I feel like single-letter params may be confusing on command line, since they are also used as shortcuts for longer names. Should we consider something like "completions", "n_completions", or "completions_count"?

docsite/posts/config/env_vars.md
darthtrevino: If we're pre-1.0, then we can release this as a minor version bump. The fallbacks are useful for expected env-vars like `OPENAI_API_KEY`, but for stuff like this I think we should just change it over. The fallbacks introduce some complexity.

docsite/posts/config/json_yaml.md
AlonsoGuevara: Sure, I can change to a more lengthy description. That said, whenever we build the dictionary it will need to be "n" as this is the parameter expected by the LLM API

docsite/posts/config/env_vars.md
AlonsoGuevara: And, another thing here is that RPM and TPM were never working, because it was named in one way and parsed in another one, so I don't thing we would need a fallback.

docsite/posts/config/json_yaml.md
natoverse: I see - in that case it is probably less confusing to align with the LLM API

AlonsoGuevara: Oh btw, this isn't used on cli, just on config. And, if stored in a env var, i would be named `GRAPHRAG_LLM_N`

PR #521, opened 2024-07-12
Fixed graphrag.query llm parameter settings not passing through to llm call. 

zanderjiang:
## Description

[Provide a brief description of the changes made in this pull request.]

## Related Issues

## Proposed Changes

## Checklist

## Additional Notes

The main change is to the global_search_config_input.py and local_search_config_input.py files to accommodate additional llm parameters (temperature, top_p, n). The current branch assumes that all llm calls are classed under llm, so no dedicated global or local llm parameters. If go that route, change the Sections.llm in create_graph_config.py.


zanderjiang:

graphrag/config/create_graphrag_config.py
```diff
@@ -480,9 +480,15 @@ def hydrate_parallelization_params(
 
         with (
             reader.use(values.get("local_search")),
-            reader.envvar_prefix(Section.local_search),
+            reader.envvar_prefix(Section.llm),
```
KylinMountain, line 483: why do you reading llm section environment? 
This will drop the original param like community_prop, top_k_entities and etc. 
I think it's not good.

zanderjiang: Yes I made a comment about that. Currently there is no dedicated variable to set the temperature, top_p, and n parameters for a local/global search call. It should be changed back to local_search when the pr for that is merged.

zanderjiang: I set it to llm to do local testing, otherwise the parameters would be None and the call will be passed with the default params.

PR #530, opened 2024-07-12
Fix/few shot selection

j2whiting:
## Description

This PR enforces a minimum of 3 few shot examples used in the final entity & relation extraction prompts. It also adds a KNN, content-based selection method for few shot examples to prevent edge cases where random sampling selects noisy, non-representative examples.

## Proposed Changes

* Prompt tune user supplied argument `--min-examples-required` which controls the number examples used in generator/entity_extraction_prompt.py. Defaults to 2.
* Added new sampling method in loader/input.py
* I've renamed some objects to reflect the change from `finetune` -> `prompt_tune`

## Additional Notes

Research required for additional improvements:

We could **theoretically** fit some clusters over the embedding space and sample from each cluster independently if we suspect there are independent distributions in the space worth sampling from independently. We would have to make some strong assumptions about what constitutes a "topic" and the geometry of topics in embedding space, which makes this a challenging problem.


graphrag/prompt_tune/generator/entity_extraction_prompt.py
```diff
@@ -79,8 +80,8 @@ def create_entity_extraction_prompt(
 
         example_tokens = num_tokens_from_string(example_formatted, model=encoding_model)
 
-        # Squeeze in at least one example
-        if i > 0 and example_tokens > tokens_left:
+        # Ensure at least three examples are included
```
AlonsoGuevara, line 83: 3 examples will potentially generate a huge prompt of over 2000 tokens. I was tying to keep it below that as it is the recommended input size by OpenAI.
Did you see any drawback in your experiments? How big is the resulting prompt with 3 examples?

j2whiting: In my most recent tests I was only receiving 1-2 few shot examples from prompt tuning. I am more or less trying to align the generated prompts with the human created ones, all of which have 3 few shot examples AFAIK. 

I do see the concern here though. Those 2000 tokens will be fed in to the LLM for each entity/relationship extraction call.

Perhaps we can set the few shot examples to collectively have a maximum length, and truncate them if they exceed that length? What do you think?

AlonsoGuevara: Wonder if that could possible confuse the algorithm given that we are truncating the examples... May be worth to test it a bit and see :)

j2whiting: I wouldn't truncate the generated entity/relationship lines. I would just truncate the examples before are sent to `ENTITY_RELATIONSHIPS_GENERATION_JSON_PROMPT`. For example, we could truncate them to 200 tokens per input example, as found here: `graphrag/index/graph/extractors/graph/prompts.py`.


The prompt at `graphrag/index/graph/extractors/graph/prompts.py` contains `2212` using `cl100k_base`.

Testing the updated prompt_tune on two of my datasets, one short noisy comms data and another rich PDF documentation we get:

```
run #1: 1807
run #2: 2076
run #3 (rich PDF text): 2892
run #4 (rich PDF text): 2155
```

Seems in line with the length of the demo prompt. Perhaps we should add an arg that allows the user to select the minimum few shot examples? I am certain that open source models used in a few-shot setting will need more examples than gpt4o for instance.

AlonsoGuevara: Sure! Let's add that! run#4 looks fine, run#3 is on the big end and if it becomes the norm we may end up having difficulty handling it on different models.

AlonsoGuevara: @j2whiting
 
 Aiming for a new version release today. Can you update this PR please?

j2whiting: Yep. Let me update branch and clean this up.

j2whiting:
@AlonsoGuevara Currently this test is failing which seems unrelated to the PR

```
FAILED tests/unit/config/test_default_config.py::TestDefaultConfig::test_default_config_with_oai_key_passes_obj - graphrag.config.errors.AzureApiBaseMissingError: API Base is required for Completion API. Please set either the OPENAI_API_BASE, GRAPHRAG_API_BASE or GRAPHRAG_API_BASE environment variable.
FAILED tests/unit/indexing/test_init_content.py::test_init_yaml - graphrag.config.errors.ApiKeyMissingError: API Key is required for Completion API. Please set either the OPENAI_API_KEY, GRAPHRAG_API_KEY or GRAPHRAG_LLM_API_KEY environment variable.
FAILED tests/unit/indexing/test_init_content.py::test_init_yaml_uncommented - graphrag.config.errors.ApiKeyMissingError: API Key is required for Completion API. Please set either the OPENAI_API_KEY, GRAPHRAG_API_KEY or GRAPHRAG_LLM_API_KEY environment variable.
```

graphrag/prompt_tune/generator/entity_extraction_prompt.py
j2whiting: Updated so the user can select the value they want, with. a default value of 2 few shot examples.

AlonsoGuevara (approved):

Issue #345, opened 2024-07-03
[Ollama] GraphRAG Community Support for running Ollama


dx111ge:
is there a working example for using Ollama? Or is it not supposed to work? Did try, but without any success. 

Thanks in advance

bmaltais:
Embeddings are not working with Olamma... I was able to get things working with Ollama for the entities and openai for embeddings.

bmaltais:
Working config can be found here: https://github.com/microsoft/graphrag/issues/339#issuecomment-2206149531

av:
Ollama works as expected

```bash
GRAPHRAG_API_KEY=123
GRAPHRAG_API_BASE=http://172.17.0.1:11434/v1
# GRAPHRAG_LLM_MODEL=llama3:instruct
GRAPHRAG_LLM_MODEL=codestral
GRAPHRAG_LLM_THREAD_COUNT=4
GRAPHRAG_LLM_CONCURRENT_REQUESTS=8
GRAPHRAG_LLM_MAX_TOKENS=2048

GRAPHRAG_EMBEDDING_API_BASE=http://172.17.0.1:11435/v1
GRAPHRAG_EMBEDDING_MODEL=mxbai-embed-large
```

`:11435` is a dead-simple proxy that converts HTTP requests from OAI to Ollama format

<details><summary>API shapes</summary>
<p>

### OAI
```typescript
JSON.stringify({
  object: "list",
  data: [
    ...results.map((r, i) => ({
      object: "embedding",
      index: i,
      embedding: r.embedding,
    })),
  ],
  model,
  usage: {
    prompt_tokens: 0,
    total_tokens: 0,
  },
})
```

### Ollama
```typescript
JSON.stringify({
  model,
  prompt: input,
})
```

</p>
</details>

bmaltais:
> Ollama works as expected
> 
> ```shell
> GRAPHRAG_API_KEY=123
> GRAPHRAG_API_BASE=http://172.17.0.1:11434/v1
> # GRAPHRAG_LLM_MODEL=llama3:instruct
> GRAPHRAG_LLM_MODEL=codestral
> GRAPHRAG_LLM_THREAD_COUNT=4
> GRAPHRAG_LLM_CONCURRENT_REQUESTS=8
> GRAPHRAG_LLM_MAX_TOKENS=2048
> 
> GRAPHRAG_EMBEDDING_API_BASE=http://172.17.0.1:11435/v1
[...]

PR #551, opened 2024-07-14
JSON parsing: always fix all incoming json when using _manual_json

MarkJGx:
## Description

This change set adds non-LLM-based JSON malformity handling as a preliminary step before using the more resource-intensive LLM-based fixup.

### More Details

While running GraphRAG with a local Ollama model, I noticed frequent malformed JSON responses from LLM requests, significantly slowing down the process on an M1 Max MacBook. In a fast, parallel cloud inference system, this issue is manageable, but locally it becomes a bottleneck. After indexing, I found 140 instances of JSON parsing failures.

The `json_repair` library effectively fixed the malformed JSON in my tests. I opted not to delve into the specific parsing failure cases, as they are mainly LLM-related and predicting every edge case is impractical. This library should be robust enough to handle most local LLM faults.

## Related Issues

- https://github.com/microsoft/graphrag/issues/345

## Proposed Changes

- Add `json_repair` as a new Poetry dependency.
  - https://pypi.org/project/json-repair/
- Use `json_repair` for initial JSON repair in `_manual_json`, with a fallback to the LLM.
- Apply JSON repair when graph search JSON parsing fails.

## Checklist

- [x] Tested these changes locally.
- [x] Reviewed the code changes.
- [x] Updated documentation (if necessary).
- [ ] Added appropriate unit tests (if applicable).

### DIFF
--- a/graphrag/llm/openai/openai_chat_llm.py
+++ b/graphrag/llm/openai/openai_chat_llm.py
@@ -16,7 +16,7 @@ from graphrag.llm.types import (
     LLMOutput,
 )
 
-from ._json import clean_up_json
+from ._json import clean_up_json, fix_malformed_json
 from ._prompts import JSON_CHECK_PROMPT
 from .openai_configuration import OpenAIConfiguration
 from .types import OpenAIClientTypes
@@ -120,6 +120,7 @@ class OpenAIChatLLM(BaseLLM[CompletionInput, CompletionOutput]):
         result = await self._invoke(input, **kwargs)
         history = result.history or []
         output = clean_up_json(result.output or "")
+        output = fix_malformed_json(result.output or "")
         try:
             json_output = try_parse_json_object(output)
             return LLMOutput[CompletionOutput](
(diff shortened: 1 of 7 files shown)


MarkJGx:

graphrag/llm/openai/openai_chat_llm.py
```diff
@@ -120,6 +120,7 @@ async def _manual_json(
         result = await self._invoke(input, **kwargs)
         history = result.history or []
         output = clean_up_json(result.output or "")
+        output = fix_malformed_json(result.output or "")
```
s106916, line 123: output = fix_malformed_json(output)

MarkJGx: Your comment seems to have gotten messed up, but I did notice the issue with the output not being forwarded from one function another. Just addressed.

MarkJGx:
Hey. I've addressed the oversight caught in the comment and rebased on main, while rehashing the poetry lock with the new library added. Waiting for a review. @AlonsoGuevara @jgbradley1 et al.

natoverse:
We have resolved several issues related to text encoding and JSON parsing that are rolled up into version [0.2.2](https://github.com/microsoft/graphrag/releases/tag/v0.2.2). Please try again with that version and re-open if this is still an issue.

Issue #575, opened 2024-07-16
[Issue]: All workflows completed successfully，but graphrag failed to answer any question given the provided data


YP-Yang:
### Describe the issue

I use local LLM to run graphrag and completed all workflows successfully, but graphrag failed to answer any question given the provided data

I found there's an error like below, don't know if it is the cause
RuntimeError: Failed to generate valid JSON output
00:33:57,636 graphrag.index.reporting.file_workflow_callbacks INFO Community Report Extraction Error details=None
00:33:57,636 graphrag.index.verbs.graph.report.strategies.graph_intelligence.run_graph_intelligence WARNING No report found for community: 0

my local LLM is llama3 in ollama and nomic-embed-text-v1.5.Q5_K_M.gguf in LM-studio

the input file book.txt is just the same as in the GraphRAG Get Started doc: 
curl https://www.gutenberg.org/cache/epub/24022/pg24022.txt > ./ragtest/input/book.txt

### GraphRAG Config Used

this is my .env file
GRAPHRAG_API_KEY=ollama

this is my setting.yaml file
![image](https://github.com/user-attachments/assets/52a3640d-8013-48b4-aa4e-68fcc1c83475)

### Logs and screenshots

this is the terminal output
(GraphRAG) D:\AI\RAG\GraphRAG>python -m graphrag.index --root .
🚀 Reading settings from settings.yaml
D:\anaconda3\envs\GraphRAG\Lib\site-packages\numpy\core\fromnumeric.py:59: FutureWarning: 'DataFrame.swapaxes' is
deprecated and will be removed in a future version. Please use 'DataFrame.transpose' instead.
  return bound(*args, **kwds)
🚀 create_base_text_units
                                   id  ... n_tokens
0    680dd6d2a970a49082fa4f34bf63a34e  ...      300
1    95f1f8f5bdbf0bee3a2c6f2f4a4907f6  ...      300
2    3a450ed2b7fb1e5fce66f92698c13824  ...      300
3    95b143eba145d91eacae7be3e4ebaf0c  ...      300
4    c390f1b92e2888f78b58f6af5b12afa0  ...      300
..                                ...  ...      ...
226  972bb34ddd371530f06d006480526d3e  ...      300
227  2f918cd94d1825eb5cbdc2a9d3ce094e  ...      300
228  eec5fc1a2be814473698e220b303dc1b  ...      300
229  535f6bed392a62760401b1d4f2aa5e2f  ...      300
[...]

Issue #528, opened 2024-07-12
[Bug]: <title>Error embedding chunk {'OpenAIEmbedding': "'NoneType' object is not iterable"}


hongyispace:
### Describe the bug

I have finished the pipeline with llm and ebedding of ollama. However, when I tried to query: python -m graphrag.query --root pg18v8 --method local "describe the types of impact craters"

The error message is:
INFO: Reading settings from pg18v8/settings.yaml
creating llm client with {'api_key': 'REDACTED,len=56', 'type': "openai_chat", 'model': 'llama3:70b-instruct-q5_K_M', 'max_tokens': 4000, 'request_timeout': 180.0, 'api_base': 'http://localhost:11434/v1', 'api_version': None, 'organization': None, 'proxy': None, 'cognitive_services_endpoint': None, 'deployment_name': None, 'model_supports_json': True, 'tokens_per_minute': 0, 'requests_per_minute': 0, 'max_retries': 1, 'max_retry_wait': 10.0, 'sleep_on_rate_limit_recommendation': True, 'concurrent_requests': 1}
creating embedding llm client with {'api_key': 'REDACTED,len=56', 'type': "openai_embedding", 'model': 'nomic-embed-text', 'max_tokens': 4000, 'request_timeout': 180.0, 'api_base': 'http://localhost:11434/api', 'api_version': None, 'organization': None, 'proxy': None, 'cognitive_services_endpoint': None, 'deployment_name': None, 'model_supports_json': None, 'tokens_per_minute': 0, 'requests_per_minute': 0, 'max_retries': 3, 'max_retry_wait': 10.0, 'sleep_on_rate_limit_recommendation': True, 'concurrent_requests': 25}
Error embedding chunk {'OpenAIEmbedding': "'NoneType' object is not iterable"}
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/opt/miniconda3/envs/grag/lib/python3.12/site-packages/graphrag/query/__main__.py", line 75, in <module>
    run_local_search(
  File "/opt/miniconda3/envs/grag/lib/python3.12/site-packages/graphrag/query/cli.py", line 154, in run_local_search
    result = search_engine.search(query=query)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #609, opened 2024-07-18
Fix: bugs when using opensource models

PaulSZH95:
## Description

Bugs fixes

## Related Issues

#575 #528

## Proposed Changes

1 - clean_up_json function is resolved to parse from first instance of '{' in llm outputs. Allows for opensource models which are more verbose.

2 - embed function resolves to decode encoded chunked tokens. This allows opensource models to with a different tokenizer to still work.

## Checklist

## Additional Notes

I have not tested with openai's model, only with llm: groq and llm. embedding: lmstudio.

### DIFF
--- a/graphrag/index/utils/json.py
+++ b/graphrag/index/utils/json.py
@@ -6,6 +6,7 @@
 
 def clean_up_json(json_str: str):
     """Clean up json string."""
+    json_str = json_str[json_str.index('{'):]
     json_str = (
         json_str.replace("\\n", "")
         .replace("\n", "")
--- a/graphrag/query/llm/oai/embedding.py
+++ b/graphrag/query/llm/oai/embedding.py
@@ -82,6 +82,7 @@ class OpenAIEmbedding(BaseTextEmbedding, OpenAILLMImpl):
         chunk_lens = []
         for chunk in token_chunks:
             try:
+                chunk = self.token_encoder.decode(chunk)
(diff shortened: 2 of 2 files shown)


PaulSZH95:

graphrag/index/utils/json.py
```diff
@@ -6,6 +6,7 @@
 
 def clean_up_json(json_str: str):
     """Clean up json string."""
+    json_str = json_str[json_str.index('{'):]
```
cedarwud, line 9: it will raise error in global query

json_str = json_str[json_str.index('{'):]
ValueError: substring not found

PaulSZH95: wrt to your error... no matter how well you write your json parser you'd still encounter error from time to time. 

reason: your model isn't able to output json in a format you require it to.

So far the only solution is reiteration when errors are occured as error will still occur from time to time even if you finetune your gpt4 model. Of course, this is speaking from experience but so far i have not seen any model getting a 100% in humaneval benchs.

Probably a good fix would be reiteration when faced with parsing error and not a fix on the parsing logic.

natoverse:
We have resolved several issues related to text encoding and JSON parsing that are rolled up into version [0.2.2](https://github.com/microsoft/graphrag/releases/tag/v0.2.2). Please try again with that version and re-open if this is still an issue.

(this may not resolve embeddings formats, but our expectation is that any proxy will translate to maintain compatibility with the default GraphRAG LLM calls)

PR #642, opened 2024-07-21
Update community_reports_extractor.py to accept integer rating

wenwkich:
## Description

There were "Failed to generate valid JSON output" errors when generating community reports, it turns out it could be because some models (I'm using Ollama) output integer ratings for the results, but the code validating the result requires rating to be a float.

## Proposed Changes

Update code community_reports_extractor.py to let the `is_response_valid` lambda to accept `float | int` instead of just `float`

## Checklist


danielcmm:
Needed

AlonsoGuevara:
Hi!
We have addressed this in #932
Closing this, thanks a lot for your contribution

Issue #657, opened 2024-07-22
Support model providers other than OpenAI and Azure


natoverse:
### Is your feature request related to a problem? Please describe.

Right now GraphRAG only natively supports models hosted by OpenAI and Azure. Many users would like to run additional models, including alternate APIs, SLMs, or models running locally. As a research team with limited bandwidth it is unlikely we will add native support for more model providers in the near future. Our focus is on memory structures and algorithms to improve LLM information retrieval, and we've got a lot of experiments in the queue!

There are alternative options to achieve extensibility, and many GraphRAG users have had luck extending the library. So far we've seen this most commonly with Ollama, which runs on localhost and supports a very wide variety of models. This approach depends on Ollama supporting the standard OpenAI API for chat completion and embeddings so it can proxy our API calls, and it looks like this is working for a lot of folks (though may require some hacking).

*Please note*: while we are excited to see GraphRAG used with more models, our team will not have time to help diagnose issues. We'll do our best to route bug reports to existing conversations that might be helpful. For the most part you should expect that if you file a bug related to running an alternate solution, we'll link to this issue, a relevant conversation if we're aware of one, and then we'll close the bug.

Here is a general discussion regarding OSS LLMs: https://github.com/microsoft/graphrag/discussions/321.

And a couple of popular Ollama-related issues: #339 and #345. We'll link to others in the comments when relevant.

Have a look at issues tagged with the [community_support](https://github.com/microsoft/graphrag/labels/community_support) label as well.

natoverse:
Here is a user-submitted solution for fixing the community rating value. We normally expect a float, but some models will send an int. https://github.com/microsoft/graphrag/issues/772

nileshtrivedi:
If GraphRAG start's using Microsoft's own agent library [AutoGen](https://github.com/microsoft/autogen), supporting different model providers would not need to be re-implemented here.

[LLM Configurations docs for AutoGen](https://microsoft.github.io/autogen/docs/topics/llm_configuration)

natoverse:
@Mxk-1 has found chunking settings that help resolve issues with create_base_entity_graph with Ollama:

>>>The chunk splitting in the original setting.yaml provided may not be suitable for the model launched with Ollama, as it could be either too large or too small, leading to errors in the model's responses. The original paper mentioned using the GPT-4o model, while the model I deployed locally is Gemma2:9b via Ollama. These two models differ in size and performance.

[...]

PR #738, opened 2024-07-26
Update issue templates for more explicit guidance

natoverse:
## Description

Updated the issue templates to try and avoid things that have already been addressed


natoverse:
FYI I removed the direct reference to #657 because GitHub auto-links every new issue to it, which is noisy

PR #665, opened 2024-07-23
Two curly brackets to one curly bracket

jiangbo721:
When I use a single curly brace the returned json string is more accurate than two curly braces

## Description

Two curly brackets to one curly bracket

## Checklist

- [x] I have tested these changes locally.
- [x] I have reviewed the code changes.
- [ ] I have updated the documentation (if necessary).
- [ ] I have added appropriate unit tests (if applicable).

PR #652, opened 2024-07-22
Use clean_up_json to process strings when executing try_parse_json_ob…

jiangbo721:
[Reference any related issues or tasks that this pull request addresses.]

## Proposed Changes

Solved the case where try_parse_json_object handles json strings in markdown or other formats that can be solved by letting the application return normally.

## Checklist

- [x] I have tested these changes locally.
- [x] I have reviewed the code changes.
- [ ] I have updated the documentation (if necessary).
- [ ] I have added appropriate unit tests (if applicable).

## Additional Notes

[Add any additional notes or context that may be helpful for the reviewer(s).]

PR #617, opened 2024-07-19
Json validation and fix

s106916:
## Description
1.  This PR consolidates json string validation and loading from index, report and llm three parts into one piece.
2.  including trim extract description from llm result, replace or convert unfit chars, using json_repair for the finally repair before json loading. wrapped all functions into try_parse_json_object() as unique one function to use.     
3. There are two duplicated json_clean_up functions in both index and llm section, consolidate into one under llm model.
4. Furthermore, wrap json_clean_up into try_parse_json_object() function without Exception for friendly retry 

## Proposed Changes
add json-repair

## Checklist

- [x] I have tested these changes locally.
- [x] I have reviewed the code changes.
- [x] I have updated the documentation (if necessary).
- [ ] I have added appropriate unit tests (if applicable).

## Additional Notes

[Add any additional notes or context that may be helpful for the reviewer(s).]

PR #801, opened 2024-08-02
Repair json when LLM returns faulty responses on non json mode

AlonsoGuevara:
## Description
1.  This PR consolidates json string validation and loading from index, report and llm three parts into one piece.
2.  including trim extract description from llm result, replace or convert unfit chars, using json_repair for the finally repair before json loading. wrapped all functions into try_parse_json_object() as unique one function to use.     
3. There are two duplicated json_clean_up functions in both index and llm section, consolidate into one under llm model.
4. Furthermore, wrap json_clean_up into try_parse_json_object() function without Exception for friendly retry 

## Related PRs
#617 (This is a merge of this) kudos @s106916 
#665
#652 
#551 

## Proposed Changes
add json-repair

## Checklist

## Additional Notes


natoverse (approved):

s106916:
I will close my one. :) #https://github.com/microsoft/graphrag/pull/617

xxll88:
although json_clean_up repair json faulty responses，it increase index time and global search time 70-80%
v0.2.0 index time 35min  , v0.2.1   62min
v0.2.0 global search time  45s , v0.2.1   75s

graphrag/llm/openai/utils.py
```diff
@@ -87,17 +89,51 @@ def get_completion_llm_args(
     }
 
 
-def try_parse_json_object(input: str) -> dict:
-    """Generate JSON-string output using best-attempt prompting & parsing techniques."""
+def try_parse_json_object(input: str) -> tuple[str, dict]:
```
jgbradley1, line 92: @AlonsoGuevara there is one edge case here that may explain @xxll88's comment about the significant increase in indexing/query time. Assume `input` was json-formatted correctly by the LLM - the logic in this function would still attempt to call `.replace()` multiple times on the string when it's not necessary.

Perhaps there should be a try-except call made to `json.loads(...)` at the top of this function to first check whether any further string operations are needed. Arguably, this *edge case* is the most generic case encountered when using a model that guarantee's json output. It turns into a lot of unnecessary string operations.

MarkJGx:
No attribution to my original PR which still hasn't received a review??
https://github.com/microsoft/graphrag/pull/551

graphrag/llm/openai/utils.py
AlonsoGuevara: Thanks for your comments @jgbradley1 and @xxll88 
Addressing this on #834

PR #839, opened 2024-08-06
Implement query api

jgbradley1:
## Description

This PR cleans up the code between the CLI and the underlying library for the query engine and introduces the initial start to an API design that other systems/users may use to interface with graphrag programmatically.

## Checklist


graphrag/query/api.py
```diff
@@ -0,0 +1,200 @@
...
+            )
+        )
+
+    return description_embedding_store
+
+
+@validate_call(config={"arbitrary_types_allowed": True})
+def global_search(
+    config: GraphRagConfig,
+    final_nodes: pd.DataFrame,
```
natoverse, line 88: "final" throughout here may be helpful for clarity due to our parquet naming convention (i.e., based on the step name), but it seems a bit specific as an API. "nodes", "entities" etc. is more semantically correct. Thoughts?

jgbradley1: I agree. I removed "final" from the variable names but left it in the docstring for documentation purposes.

AlonsoGuevara (approved):

AlonsoGuevara (approved):

PR #844, opened 2024-08-06
community_report_issue_slove

prasantpoudel:
1 .When the community report are generated, some time model give the rating in int format, but in code it accept as float.
2. When the community report are generated and parse by the code then due to "{{" throw the error like error in parsing the JSON. Converting the "{{" to "{" in the community report solves the error. And in the community report the required structure and # Grounding Rules are written 2 time. This is not necessary, we can delete the below one, this help to reduce the token size problem if we are running the LLMs model Locally.

## Checklist


prasantpoudel:

prasantpoudel:

natoverse:
We have a number of JSON formatting issues that are resolved with release [0.2.2](https://github.com/microsoft/graphrag/releases/tag/v0.2.2). This may resolve the issues you are trying to address. Note that we do need the doubled braces because we do variable replacement.

PR #882, opened 2024-08-09
Added streaming output support for local search & global search

6ixGODD:
## Description

This PR introduces changes to enable streaming output in the LLM's generate method, as well as in the search methods of both `LocalSearch` and `GlobalSearch` classes. Intended to facilitate web application development or any other scenarios where streaming and real-time display of generated content is required.

## Related Issues

none

## Proposed Changes

1. Updated the `generate` method's return type to `Generator[str, None, str]` of `BaseLLM` and `ChatOpenAI` and implement it.
2. Updated the `search` method's return type to `Generator[str, None, SearchResult]` of `LocalSearch`, `GlobalSearch` and `BaseSearch`, and implement it. 
3. Added a '--streaming' flag to the cli, allows users to specify whether they want to enable streaming output for local and global search.

## Checklist

## Additional Notes

A caveat with these changes is that when a method contains a yield statement, the Python interpreter automatically defines it as a generator function, meaning that it always returns a generator. If streaming is not needed, it becomes less straightforward to directly retrieve the final result, as you must use a try-except block to catch the StopIteration exception and extract the value from it. This can lead to less elegant code.

If necessary, the streaming functionality could be refactored into separate methods to maintain clarity and separation of concerns.

### DIFF
--- a/graphrag/query/__main__.py
+++ b/graphrag/query/__main__.py
@@ -73,6 +73,12 @@ if __name__ == "__main__":
     )
 
     parser.add_argument(
+        "--streaming",
+        help="Whether to output the response text in a streaming (chunk-by-chunk) format",
+        action="store_true",
+    )
+
+    parser.add_argument(
         "query",
         nargs=1,
         help="The query to run",
@@ -89,6 +95,7 @@ if __name__ == "__main__":
                 args.root,
                 args.community_level,
                 args.response_type,
+                args.streaming,
                 args.query[0],
             )
         case SearchType.GLOBAL:
@@ -98,6 +105,7 @@ if __name__ == "__main__":
                 args.root,
                 args.community_level,
                 args.response_type,
+                args.streaming,
                 args.query[0],
             )
         case _:
--- a/graphrag/query/api.py
+++ b/graphrag/query/api.py
@@ -90,6 +90,7 @@ async def global_search(
     community_reports: pd.DataFrame,
     community_level: int,
     response_type: str,
+    streaming: bool,
     query: str,
 ) -> str | dict[str, Any] | list[dict[str, Any]]:
     """Perform a global search.
@@ -120,10 +121,28 @@ async def global_search(
         entities=_entities,
         response_type=response_type,
     )
-    result = await search_engine.asearch(query=query)
-    reporter.success(f"Global Search Response: {result.response}")
-    return result.response
 
+    if not streaming:
+        result = await search_engine.asearch(query=query)
+        reporter.success(f"Global Search Response: {result.response}")
+        return result.response
+
+    else:
+        import sys
+        full_resp = ''
+        results = search_engine.astream_search(query=query)
+        reporter.success(f'Global Search Response: \n')
+        sys.stdout.write('⬤')  # display an initial progress indicator
+
+        async for result in results:
+            sys.stdout.write('\b')  # move the cursor back one position to overwrite the previous progress indicator
+            sys.stdout.flush()  # flush the output buffer
+            sys.stdout.write(result + "⬤")  # display the next progress indicator
+            sys.stdout.flush()
+            full_resp += result
+
+        sys.stdout.write('\b\n')
+        return full_resp
 
 @validate_call(config={"arbitrary_types_allowed": True})
 async def local_search(
@@ -136,6 +155,7 @@ async def local_search(
     covariates: pd.DataFrame | None,
     community_level: int,
     response_type: str,
+    streaming: bool,
     query: str,
 ) -> str | dict[str, Any] | list[dict[str, Any]]:
     """Perform a local search.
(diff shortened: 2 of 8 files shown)


6ixGODD:
> @6ixGODD please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @microsoft-github-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement

6ixGODD:


Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

jgbradley1:
@6ixGODD This is great work. You caught us in the middle of defining an API layer between the CLI and library code. We just merged a [PR](https://github.com/microsoft/graphrag/pull/839) that more explicitly defines the query API.

I agree with the separation of concerns issue. Can you update/refactor to use this existing query API and if there needs to be separate functions to maintain clarity, we're open to adding new streaming functions to the API.

6ixGODD:
> @6ixGODD This is great work. You caught us in the middle of defining an API layer between the CLI and library code. We just merged a [PR](https://github.com/microsoft/graphrag/pull/839) that more explicitly defines the query API.
> 
> I agree with the separation of concerns issue. Can you update/refactor to use this existing query API and if there needs to be separate functions to maintain clarity, we're open to adding new streaming functions to the API.

I've made the corresponding changes:

I kept the --streaming command-line argument, which controls whether the output is streamed.
I modified the two APIs in api.py by adding conditional logic to decide whether to use streaming output, and included a ⬤ as a progress indicator (free to remove it if it looks out of place).
I added stream_generate and astream_generate methods to the BaseLLM abstract class for synchronous and asynchronous streaming response generation, respectively, and implemented them in the ChatOpenAI class (I also added a few private methods, following the original code style).
I added an astream_search method to the BaseSearch abstract class for streaming search, and implemented it in both GlobalSearch and LocalSearch classes.
I made the necessary updates in the CLI to support the --streaming option.
With these changes, apart from adding the streaming parameter and a few additional methods, I have made minimal modifications to the original code.

KylinMountain:
there's on_llm_new_token callback, why don't you use this method?

6ixGODD:
> there's on_llm_new_token callback, why don't you use this method?

@KylinMountain I missed the `on_llm_new_token` callback😂 I check the code and youre right. But from an application development perspective, whether using websocket, sse, or others, typically needs to be a way to YIELD chunks of data.. I figured why not just handle it directly in the class instead of decoupling it. right?

jgbradley1:
@6ixGODD is right. The callback approach does work but it's quite messy for an external system to use and could be prone to errors that you might not think about (when considering dropped messages/tokens) depending on the use case you're targeting.

For example, in the solution accelerator, we had to [implement some custom data structures and logic](https://github.com/Azure-Samples/graphrag-accelerator/blob/main/backend/src/api/experimental.py) in order to create a proper _streaming_ web API that would work like most programmatic clients expect (i.e. using the `requests` library to iterate over a response chunks).

Adding a generator function to the API is much cleaner for everyone to use (my opinion). I copied this fork and made a [new PR](https://github.com/microsoft/graphrag/pull/944) (under review) that cleaned up several other things that we needed to take care of in order to expose a proper streaming API as a generator function.

6ixGODD:
> @6ixGODD is right. The callback approach does work but it's quite messy for an external system to use and could be prone to errors that you might not think about (when considering dropped messages/tokens) depending on the use case you're targeting.
> 
> 
> 
> For example, in the solution accelerator, we had to [implement some custom data structures and logic](https://github.com/Azure-Samples/graphrag-accelerator/blob/main/backend/src/api/experimental.py) in order to create a proper _streaming_ web API that would work like most programmatic clients expect (i.e. using the `requests` library to iterate over response chunks).
> 
> 
> 
> Adding a generator function to the API is much cleaner for everyone to use (my opinion). I copied this fork and made a [new PR](https://github.com/microsoft/graphrag/pull/944) (under review) that cleaned up several other things that were needed in order to expose a proper streaming API as a generator function.

@jgbradley1 Completely agree😄I saw your new PR, and I am honored to have the possibility of contributing to this repository.

Issue #891, opened 2024-08-11
[Bug]: NotADirectoryError: [Errno 20] Not a directory: '/Users/username/ragtest/output/.DS_Store/artifacts/create_final_nodes.parquet'


semenoffalex:
### Do you need to file an issue?

- [X] I have searched the existing issues and this bug is not already filed.
- [ ] My model is hosted on OpenAI or Azure. If not, please look at the "model providers" issue and don't file a new one here.
- [X] I believe this is a legitimate bug, not just a question. If this is a question, please use the Discussions area.

### Describe the bug

I managed to follow your example here and got the message: "All workflows completed successfully", even though I saw "Errors occurred during the pipeline run, see logs for more details" message couple of times at the "create_base_entity_graph" step.

However, when I run the first command for interaction with the graph I got (it took my M2 ~ 5 hours to create it for "A Christmas Carol") I got some Python error.

Command I'm trying:
```
python -m graphrag.query \
--root ./ragtest \
--method global \
"What are the top themes in this story?"
```

Error I get:
```
INFO: Reading settings from ragtest/settings.yaml
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/opt/miniconda3/envs/graphollama/lib/python3.11/site-packages/graphrag/query/__main__.py", line 83, in <module>
    run_global_search(
  File "/opt/miniconda3/envs/graphollama/lib/python3.11/site-packages/graphrag/query/cli.py", line 67, in run_global_search
    final_nodes: pd.DataFrame = pd.read_parquet(
                                ^^^^^^^^^^^^^^^^
  File "/opt/miniconda3/envs/graphollama/lib/python3.11/site-packages/pandas/io/parquet.py", line 667, in read_parquet
    return impl.read(
           ^^^^^^^^^^
  File "/opt/miniconda3/envs/graphollama/lib/python3.11/site-packages/pandas/io/parquet.py", line 267, in read
    path_or_handle, handles, filesystem = _get_path_or_handle(
                                          ^^^^^^^^^^^^^^^^^^^^
  File "/opt/miniconda3/envs/graphollama/lib/python3.11/site-packages/pandas/io/parquet.py", line 140, in _get_path_or_handle
    handles = get_handle(
              ^^^^^^^^^^^
  File "/opt/miniconda3/envs/graphollama/lib/python3.11/site-packages/pandas/io/common.py", line 882, in get_handle
[...]

PR #910, opened 2024-08-12
Add stricter filtering and tests for cli data directory discovery

natoverse:
## Description

Improves the filtering of folders inside the indexing output. This helps with OS-generated files or any other non-automatically-timestamped folders the user may have created.

## Related Issues

#891 

## Proposed Changes

- Uses regex match and is_dir check
- Adds unit tests for query module

## Checklist

## Additional Notes

### DIFF
--- a/graphrag/query/cli.py
+++ b/graphrag/query/cli.py
@@ -132,7 +132,7 @@ def _infer_data_dir(root: str) -> str:
     if output.exists():
         expr = re.compile(r"\d{8}-\d{6}")
         filtered = [f for f in output.iterdir() if f.is_dir() and expr.match(f.name)]
-        folders = sorted(filtered, key=os.path.getmtime, reverse=True)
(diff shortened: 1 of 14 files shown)


graphrag/query/cli.py
```diff
@@ -129,7 +129,9 @@ def _infer_data_dir(root: str) -> str:
     output = Path(root) / "output"
     # use the latest data-run folder
     if output.exists():
-        folders = sorted(output.iterdir(), key=os.path.getmtime, reverse=True)
+        expr = re.compile(r"\d{8}-\d{6}")
+        filtered = [f for f in output.iterdir() if f.is_dir() and expr.match(f.name)]
```
jgbradley1, line 133: What if you add another condition `and not f.name.startswith(".")` to guard against dot files in case the regex is not enough?

natoverse: I added another test folder to confirm that the regex is enough

AlonsoGuevara (approved):

PR #953, opened 2024-08-16
Index API

dworthen:
## Description

This PR pulls out the core library logic from the CLI layer into a dedicated API layer, useful for programmatic use cases. The CLI now calls out to the API layer. 

## Checklist

## Additional Notes


gmile:
Would it be possible to import API module into a custom python script to use it directly? If yes, could you please provide a simple example of how to do this?

(asking as barely a python expert, but interested in embedding graphrag as part of my custom python webserver)

AlonsoGuevara (approved):

graphrag/index/api.py
```diff
@@ -0,0 +1,163 @@
...
+"""
+
+import logging
+from pathlib import Path
+
+from graphrag.config.enums import CacheType, ReportingType
+from graphrag.config.models.graph_rag_config import GraphRagConfig
+from graphrag.config.resolve_timestamp_path import resolve_timestamp_path
+
+from . import create_pipeline_config as cpc
```
jgbradley1, line 18: This line seems a bit odd. Would this work (and get rid of line 30)?
```suggestion
from .create_pipeline_config import create_pipeline_config
```

dworthen: I felt the create_pipeline_config should be part of the main api but I am not using it directly in the api.py file. The single line version gets removed by our formatting and ruff rules since I am not using the method in the file and that is why I had to redeclare it. I can also remove it from the file if we don't think it needs to be exported with the main API.

AlonsoGuevara: Nice catch!! Indeed, line 300 seems weird. Removing it!

graphrag/index/api.py
```diff
@@ -0,0 +1,161 @@
...
+        log_path = resolve_timestamp_path(
+            Path(config.root_dir) / config.reporting.base_dir / "indexing-engine.log",
+            timestamp_value,
+        )
+        enable_logging(log_path, verbose)
+        return (True, str(log_path))
+    return (False, "")
+
+
+def load_progress_reporter(reporter_type: str = "none") -> ProgressReporter:
```
jgbradley1, line 85: Can this function be moved to `cli.py` as a private function`_load_progress_reporter(...)`? It doesn't look like the index API ever relies on calling this function...only the CLI does.

graphrag/index/api.py
```diff
@@ -0,0 +1,161 @@
...
+    NullProgressReporter,
+    PrintProgressReporter,
+    ProgressReporter,
+)
+from .progress.rich import RichProgressReporter
+from .run import run_pipeline_with_config
+from .typing import PipelineRunResult
+
+
+def enable_logging(log_filepath: str | Path, verbose: bool = False) -> None:
```
jgbradley1, line 31: Same comment as below. I think this function can be moved to `cli.py` as `_enable_logging(...)` as it never gets called directly by the `build_index(...)` function.

graphrag/index/api.py
jgbradley1: Disregard this comment. I see now that a user would need to call this function first in order to pass the right object into `build_index(...)`.

I think the other logging functions though are only useful to the CLI though and shouldn't be part of the API right?

dworthen: Moved the load progress code to the progress module.

graphrag/index/api.py
dworthen: Moved the logging code to be in the config module.

graphrag/index/api.py
```diff
@@ -0,0 +1,161 @@
...
+    logging.basicConfig(
+        filename=log_filepath,
+        filemode="a",
+        format="%(asctime)s,%(msecs)d %(name)s %(levelname)s %(message)s",
+        datefmt="%H:%M:%S",
+        level=logging.DEBUG if verbose else logging.INFO,
+    )
+
+
+def enable_logging_with_config(
```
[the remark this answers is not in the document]
dworthen: Moved the logging code to be in the config module.

AlonsoGuevara (approved):

caster-Q:
Is there a demo that can run this index API?

PR #996, opened 2024-08-21
Change lancedb placement

KennyZhang1:
## Description

This PR changes the placement of the lancedb vector database.

## Related Issues

N/A

## Proposed Changes

Instead of having the lancedb directory dumped into the same level as the root, the code places lancedb into the /artifacts directory of the most recent timestamp output.

## Checklist

## Additional Notes


graphrag/query/api.py
```diff
@@ -252,6 +261,12 @@ async def local_search_streaming(
     vector_store_type = vector_store_args.get("type", VectorStoreType.LanceDB)
 
     _entities = read_indexer_entities(nodes, entities, community_level)
+
+    base_dir = cast(str, root_dir) + "/" + config.storage.base_dir
```
AlonsoGuevara, line 265: I think we should use Pathlib here instead of str concat, as it ill break in Windows due to path format

graphrag/query/api.py
```diff
@@ -252,6 +261,12 @@ async def local_search_streaming(
...
+    resolved_base_dir = resolve_timestamp_path(base_dir)
+    lancedb_dir = str(resolved_base_dir) + "/lancedb"
```
AlonsoGuevara, line 267: Idem

graphrag/query/cli.py
```diff
@@ -177,12 +180,13 @@ def _configure_paths_and_settings(
...
     config_filepath: str | None,
 ) -> tuple[str, str | None, GraphRagConfig]:
+    config = _create_graphrag_config(root_dir, config_filepath)
     if data_dir is None and root_dir is None:
         msg = "Either data_dir or root_dir must be provided."
         raise ValueError(msg)
     if data_dir is None:
-        data_dir = _infer_data_dir(cast(str, root_dir))
-    config = _create_graphrag_config(root_dir, config_filepath)
+        base_dir = cast(str, root_dir) + "/" + config.storage.base_dir
```
AlonsoGuevara, line 188: +1 here, let's use Path() instead

graphrag/query/api.py
```diff
@@ -252,6 +262,11 @@ async def local_search_streaming(
     vector_store_type = vector_store_args.get("type", VectorStoreType.LanceDB)
 
     _entities = read_indexer_entities(nodes, entities, community_level)
+
+    base_dir = Path(cast(str, root_dir)) / Path(config.storage.base_dir)
+    resolved_base_dir = resolve_timestamp_path(base_dir)
+    lancedb_dir = os.path.join(resolved_base_dir, "lancedb")  # noqa: PTH118
```
AlonsoGuevara, line 268: since  resolved_base_dir is already a Path, you can just do resolved_base_dir / "lancedb"
The / operator behaves specially when working over paths and strings

AlonsoGuevara: Also same with other concats. It is not that you're doing it wrong, but you can just leverage how / behaves when one of the variables is already a Path object

graphrag/query/cli.py
```diff
@@ -177,12 +180,13 @@ def _configure_paths_and_settings(
...
 ) -> tuple[str, str | None, GraphRagConfig]:
+    config = _create_graphrag_config(root_dir, config_filepath)
     if data_dir is None and root_dir is None:
         msg = "Either data_dir or root_dir must be provided."
         raise ValueError(msg)
     if data_dir is None:
-        data_dir = _infer_data_dir(cast(str, root_dir))
-    config = _create_graphrag_config(root_dir, config_filepath)
+        base_dir = Path(cast(str, root_dir)) / config.storage.base_dir
+        data_dir = cast(str, resolve_timestamp_path(base_dir))
```
AlonsoGuevara, line 189: Instead of cast you can do str(resolve_timestamp_path(base_dir))
Generally cast is used to tell the type checker to treat the item as str, whereas str() actually converts the itme.

graphrag/query/api.py
```diff
@@ -184,6 +187,11 @@ async def local_search(
     vector_store_type = vector_store_args.get("type", VectorStoreType.LanceDB)
 
     _entities = read_indexer_entities(nodes, entities, community_level)
+
+    base_dir = Path(cast(str, root_dir)) / config.storage.base_dir
```
AlonsoGuevara, line 191: Same here, str()
Ans you'll even save an import :)

AlonsoGuevara (approved):

PR #1037, opened 2024-08-27
Query engine logger

fclarke1:
## Description

There is no query logger when using `graphrag.query.` I've implemented a query-global-engine.log and a query-local-engine.log in a very similar way to the indexing-engine.log
I found the query logger very helpful when debugging my graphrag implementation

## Proposed Changes

 - query-local-engine.log and query-global-engine.log are saved in root/output/{latest_timestamp}/reports/
 - `graphrag.query.cli` now accepts `--verbose` argument
 - query logger is now enabled using `enable_logging_with_config()` in `graphrag.query.cli`

## Checklist


fclarke1:

jgbradley1:
Closing this PR as we have since redesigned logging in graphrag to use the standard Python `logging` module.

PR #1039, opened 2024-08-27
Fix INIT_YAML embeddings default settings

TLongP:
## Description

There's a small issue with the default `settings.yaml` when initializing the workspace. In the `embeddings` configuration, `batch_size` and `batch_max_tokens` should not be part of  `llm`, causing them to always use their default values.

## Checklist


TLongP:

AlonsoGuevara (approved):

PR #1049, opened 2024-08-29
fix: Skip `resolve_timestamp_path` when using --data in local search

ka2048:
## Description
The errors during the indexing process that led to rebuilding once again have been quite frustrating, so I specified the `--resume` parameter in the CLI to avoid repeatedly consuming tokens to build the index. However, instead of using the timestamp in the example, I customized a name, like _acc_ (since there are too many previously generated timestamp directories under this directory, a custom name makes it more noticeable). After successfully completing the index build, I attempted to use global search with the `--data` parameter, and it worked well. However, when I used local search with the same `--data` parameter, I received a ValueError.

For example, the CLI command is as follows:
```python
python -m graphrag.query \
--root ./ragtest \
--data ./ragtest/output/acc/artifacts \
--method local \
"Who is Scrooge, and what are his main relationships?"
```

The ValueError originates from graphrag/config/resolve_timestamp_path.py at lines 76-78:
```python
if len(timestamp_dirs) == 0:
    msg = f"No timestamp directories found in {parent_dir} that match {pattern.pattern}."
    raise ValueError(msg)
```

This is very confusing since I didn’t use a timestamp as the value for the `--data` parameter, so why is it prompting that none were found?
Upon inspecting the code, I discovered that in the method `_configure_paths_and_settings` within the graphrag/query/cli.py file, a conditional check is already performed based on the `--data` parameter:
```python
def _configure_paths_and_settings(
    data_dir: str | None,
    root_dir: str | None,
    config_filepath: str | None,
) -> tuple[str, str | None, GraphRagConfig]:
    config = _create_graphrag_config(root_dir, config_filepath)
    if data_dir is None and root_dir is None:
        msg = "Either data_dir or root_dir must be provided."
        raise ValueError(msg)
    if data_dir is None:
        base_dir = Path(str(root_dir)) / config.storage.base_dir
        data_dir = str(resolve_timestamp_path(base_dir))
    return data_dir, root_dir, config
```

Further on, the local search methods in both graphrag/query/cli.py and graphrag/query/api.py repeatedly call `resolve_timestamp_path`, which led to the ValueError mentioned earlier.

As a result, I made modifications to the code and submitted this PR.

I modified the local_search function’s parameters to pass in the data_path parameter, which was previously handled in `_configure_paths_and_settings` in cli.py, as `resolved_base_dir`. This is only used for the `lancedb_dir` variable.

## Related Issues
I did not find any related issues.

## Checklist

---

### DIFF
--- a/.semversioner/next-release/patch-20240829071508331905.json
+++ b/.semversioner/next-release/patch-20240829071508331905.json
@@ -1,0 +1,4 @@
+{
+  "type": "patch",
+  "description": "<Skip  when using --data in local search>."
+}
--- a/graphrag/query/api.py
+++ b/graphrag/query/api.py
@@ -149,7 +149,7 @@ async def global_search_streaming(
 
 @validate_call(config={"arbitrary_types_allowed": True})
 async def local_search(
-    root_dir: str | None,
+    data_path: Path,
     config: GraphRagConfig,
     nodes: pd.DataFrame,
     entities: pd.DataFrame,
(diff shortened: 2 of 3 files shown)


ka2048:

PR #1088, opened 2024-09-03
Fix: Community Context Builder reuse reports in local mode.

jarlor:
## Description

In local mode, the community context builder will cause a report to be repeated twice, like this:
image
The elements in list `all_context_text` will be directly joined, resulting in two identical reports in the final context. This will also cause errors in the calculation of context tokens.

## Related Issues

no yet.

## Proposed Changes
- This bug is mainly caused by the failure of duplication detection of the last batch. I changed the duplicate checking method from text-based to report ID-based.
- This bug also caused the guide in the reports section of the final context to be lost, which I also fixed. 
    image

## Checklist

## Additional Notes

None


AlonsoGuevara:
Thanks @jarlor
Submitted this on #1131

PR #1090, opened 2024-09-03
Clean and organize run index code

AlonsoGuevara:
## Description

Organize the code a bit to help with incremental indexing implementation
## Related Issues

## Proposed Changes

## Checklist

## Additional Notes


graphrag/index/api.py
```diff
@@ -24,8 +22,10 @@
 
 async def build_index(
     config: GraphRagConfig,
-    run_id: str,
-    memory_profile: bool,
+    run_id: str = "",
+    is_resume_run: bool = False,
+    is_update_run: bool = False,
```
andresmor-ms, line 27: Can both `is_resume_run` and `is_update_run` be True?

graphrag/index/cli.py
```diff
@@ -113,7 +113,7 @@ def index_cli(
     """Run the pipeline with the given config."""
     progress_reporter = load_progress_reporter(reporter or "rich")
     info, error, success = _logger(progress_reporter)
-    run_id = resume or time.strftime("%Y%m%d-%H%M%S")
+    run_id = resume or update_index_id or time.strftime("%Y%m%d-%H%M%S")
```
andresmor-ms, line 116: Would be cool to add docs what happens when `resume` and `update_index_id` or if that's a mistake.

graphrag/index/run/postprocess.py
```diff
@@ -0,0 +1,50 @@
...
+    """Run the pipeline.
+
+    Args:
+        - post_process - The post process steps to run
+        - dataset - The dataset to run the steps on
+        - context - The pipeline run context
+    Returns:
+        - output - The dataset after running the post process steps
+    """
+    if post_process is not None and len(post_process) > 0:
```
andresmor-ms, line 39: I think you can replace this by just `if post_process:` since `None` and `[]` are False in python

graphrag/index/run/run.py
```diff
@@ -0,0 +1,265 @@
...
+    _validate_dataset(dataset)
+
+    log.info("Final # of rows loaded: %s", len(dataset))
+    stats.num_documents = len(dataset)
+    last_workflow = "input"
+
+    try:
+        await _dump_stats(stats, storage)
+
+        for workflow_to_run in workflows_to_run:
```
andresmor-ms, line 224: I feel like the content of this loop could go into its own function?

graphrag/index/run/run.py
```diff
@@ -0,0 +1,265 @@
...
+            # Try to flush out any intermediate dataframes
+            gc.collect()
+
+            workflow = workflow_to_run.workflow
+            workflow_name: str = workflow.name
+            last_workflow = workflow_name
```
andresmor-ms, line 230: In which case is this different to workflow_name?

graphrag/index/run/run.py
```diff
@@ -0,0 +1,265 @@
...
+        - additional_verbs - The custom verbs to use for the pipeline
+        - additional_workflows - The custom workflows to use for the pipeline
+        - debug - Whether or not to run in debug mode
+    Returns:
+        - output - An iterable of workflow results as they complete running, as well as any errors that occur
+    """
+    start_time = time.time()
+    stats = PipelineRunStats()
+    storage = storage or MemoryPipelineStorage()
+    cache = cache or InMemoryCache()
```
andresmor-ms, line 182: Could we access these 3 variables from the context declared below and move the initialization of these to the _create_run_context line? that way we remove variables from here and make sure we are using the right ones from the context

graphrag/index/run/run.py
```diff
@@ -0,0 +1,265 @@
...
+
+            log.info("Running workflow: %s...", workflow_name)
+
+            if is_resume_run and await storage.has(
+                f"{workflow_to_run.workflow.name}.parquet"
+            ):
+                log.info("Skipping %s because it already exists", workflow_name)
+                continue
+
+            stats.workflows[workflow_name] = {"overall": 0.0}
```
andresmor-ms, line 240: Probably not part of this PR but stats.workflows could be a default dict if it is always initialized this way

graphrag/index/api.py
AlonsoGuevara: At the time, yup. I was thinking adding a validation in the cli, but, makes sense to have it here as well

graphrag/index/cli.py
AlonsoGuevara: Added an exception hehe

graphrag/index/run/run.py
AlonsoGuevara: I like this idea!

graphrag/index/run/run.py
AlonsoGuevara: Removed unused var

graphrag/index/run/run.py
AlonsoGuevara: Done

graphrag/index/run/run.py
AlonsoGuevara: Done Sire

andresmor-ms (approved):

Issue #799, opened 2024-08-01
Support loading from storage blob in query library


natoverse:
### Do you need to file an issue?

- [X] I have searched the existing issues and this feature is not already filed.
- [X] My model is hosted on OpenAI or Azure. If not, please look at the "model providers" issue and don't file a new one here.
- [X] I believe this is a legitimate feature request, not just a question. If this is a question, please use the Discussions area.

### Is your feature request related to a problem? Please describe.

We have received reports of issues with blob storage (for example https://github.com/microsoft/graphrag/issues/789). It turns out that we had not yet migrated the blob loading used by the indexer into the query portion at all, so this just fails. Filing this as a "missing feature".

nociza:
Hi can you provide an example of how we can configure settings.yaml so that it can access the output folder on Azure Blob Storage? Here's my current version:

```
storage:
  type: blob
  base_dir: "output/${timestamp}/artifacts"
  connection_string: "DefaultEndpointsProtocol=https;AccountName=<my account name>;AccountKey=<my account key>;EndpointSuffix=core.windows.net"
  container_name: "data"
 ```
 
 But I'm getting "Parent directory output does not exist or is not a directory." error. 
 
 After digging into it a bit, the error seems to be arising from the resolve path function where it can't use the blob path directl y and thinks the directory doesn't exist.

dworthen:
> Hi can you provide an example of how we can configure settings.yaml so that it can access the output folder on Azure Blob Storage? Here's my current version:
> 
> ```
> storage:
>   type: blob
>   base_dir: "output/${timestamp}/artifacts"
>   connection_string: "DefaultEndpointsProtocol=https;AccountName=<my account name>;AccountKey=<my account key>;EndpointSuffix=core.windows.net"
>   container_name: "data"
> ```
> 
> But I'm getting "Parent directory output does not exist or is not a directory." error.
> 
> After digging into it a bit, the error seems to be arising from the resolve path function where it can't use the blob path directly and thinks the directory doesn't exist. Is there any way I can circumvent this? Thanks for your work! @KennyZhang1

Hi @nociza, as we discussed in #1113, we are moving away from timestamp directories. Updating your base_dir to something like "output" for both storage and reporting should resolve your issue but let us know if it does not.

nociza:
[...]

PR #1095, opened 2024-09-04
Load query from blob

KennyZhang1:
## Description

Currently, the query workflow only supports loading parquet files from local storage. This PR adds support for loading parquet files from azure blob storage.

## Related Issues

Issue #799  

## Proposed Changes

Remove hardcoding of parquet file loading from local file system at the cli level. Add a function that reads parquet files into dataframe dictionary depending on the setting defined in the config file (blob vs file loading).

## Checklist

## Additional Notes

N/A


graphrag/query/cli.py
```diff
@@ -179,3 +175,184 @@ async def run_streaming_search():
     # NOTE: we return the response and context data here purely as a complete demonstration of the API.
     # External users should use the API directly to get the response and context data.
     return response, context_data
+
+
+def _resolve_parquet_files(
```
AlonsoGuevara, line 180: Can we extract this function to an util? Can we also add an optional parameter to pass a list of the files one is interested on? 
To provide an example, I'm definitely needing this for incremental indexing, so I can load a few cherry picked outputs :)

AlonsoGuevara: Another suggestion is, the function seems a bit long, can we refactor into multiple functions?

graphrag/query/cli.py
```diff
@@ -179,3 +175,184 @@ async def run_streaming_search():
...
+
+def _create_graphrag_config(
+    root: str | None,
+    config_filepath: str | None,
+) -> GraphRagConfig:
+    """Create a GraphRag configuration."""
+    return _read_config_parameters(root or "./", config_filepath)
+
+
+def _read_config_parameters(root: str, config: str | None):
```
AlonsoGuevara, line 324: Don't we have something similar in other locations to read a config from a path?

Pretty sure it is on graphrag.config.load_config -> load_config()

AlonsoGuevara (approved):

PR #1096, opened 2024-09-04
fix setting base_dir to full paths when not using file system.

dworthen:
## Description

Do not set storage/reporting.base_dir to full paths when config type != "file". Full paths only make sense for file system based storage and logging. Blob or other remote storage/logging solutions will rely on the relative paths. 

## Checklist


graphrag/config/load_config.py
```diff
@@ -49,17 +50,20 @@ def load_config(
...
+            resolve_timestamp_path(config.reporting.base_dir, run_id)
         )
     else:
-        config.storage.base_dir = str(
-            resolve_timestamp_path((root / config.storage.base_dir).resolve())
-        )
+        config.storage.base_dir = str(resolve_timestamp_path(config.storage.base_dir))
         config.reporting.base_dir = str(
-            resolve_timestamp_path((root / config.reporting.base_dir).resolve())
+            resolve_timestamp_path(config.reporting.base_dir)
```
AlonsoGuevara, line 61: Wonder if we can simplify the login in the load_config method by translating some of the responsability to resolve_timestamp_path.
Make it more generic, something like resolve_path that receives the path, the root, the type, the id, etc. And allows for nulls and such.
Also, we can use switch case when possible, it resolves faster than if else.

KennyZhang1 (approved):
lgtm

AlonsoGuevara (approved):

PR #1113, opened 2024-09-10
Migrate towards using static output directories

dworthen:
- Fixes load_config eagering resolving directories. Directories are only resolved when the output directories are local.
- Add support for `--output` and `--reporting` flags for index CLI. To achieve previous output structure `index --output run1/artifacts --reports run1/reports`.
- Use static output directories when initializing a new project.
- Maintains backward compatibility for those using timestamp outputs locally.
- Addresses #799 

## Checklist


graphrag/config/resolve_path.py
```diff
@@ -150,3 +153,40 @@ def resolve_path(
...
+    ----------
+    config : GraphRagConfig
+        The configuration to resolve the paths in.
+    pattern_or_timestamp_value : re.Pattern[str] | str, default=None
+        The pattern to use to match the timestamp directories or the timestamp value to use.
+        If a string is provided, the path will be resolved with the given string value.
+        Otherwise, the path will be resolved with the latest available timestamp directory
+        that matches the given pattern.
+    """
+    if config.storage.type == StorageType.file:
```
AlonsoGuevara, line 177: Can we change this to a switch/case?

nociza:
Does this support using timestamp directories remotely?

dworthen:
> Does this support using timestamp directories remotely?

HI @nociza, we are migrating away from timestamp paths both remotely and locally. If you look at the v1-breaking-changes.md file in this PR it describes the change and migration. This PR still maintains support for timestamp paths locally for backwards compatibility but timestamp paths are not supported for remote storage. In issue #799 if you update your config to use static directories for storage then it should work though you may need to re-index or at least move your current index outputs to the new static location.

AlonsoGuevara (approved):

AlonsoGuevara (approved):

PR #1127, opened 2024-09-12
Cleanup cli

jgbradley1:
## Description

This PR cleans up and standardizes formatting of the various CLI. It also adds safeguards (file and directory existence checks) where appropriate.

There was one bug discovered in the `prompt_tune` module where if `--config` and `--root` were specified, the correct filepath to the config file would not get used and error out.

## Checklist

### DIFF
--- a/graphrag/index/__main__.py
+++ b/graphrag/index/__main__.py
@@ -4,15 +4,39 @@
 """The Indexing Engine package root."""
 
 import argparse
+from enum import Enum
 
 from .cli import index_cli
+
+
+class ReporterType(Enum):
+    """The type of reporter to use."""
+
+    RICH = "rich"
+    PRINT = "print"
+    NONE = "none"
+
+    def __str__(self):
+        """Return the string representation of the enum value."""
+        return self.value
+
+
+class EmitType(Enum):
+    """The type of emitter to use."""
+
+    PARQUET = "parquet"
+    CSV = "csv"
+
+    def __str__(self):
+        """Return the string representation of the enum value."""
+        return self.value
+
 
 if __name__ == "__main__":
     parser = argparse.ArgumentParser()
     parser.add_argument(
         "--config",
         help="The configuration yaml file to use when running the pipeline",
-        required=False,
         type=str,
     )
     parser.add_argument(
@@ -44,20 +68,26 @@ if __name__ == "__main__":
     )
     parser.add_argument(
         "--reporter",
-        help="The progress reporter to use. Valid values are 'rich', 'print', or 'none'",
-        type=str,
+        help="The progress reporter to use. Default='rich'.",
+        default=ReporterType.RICH,
+        type=ReporterType,
+        choices=list(ReporterType),
     )
     parser.add_argument(
         "--emit",
-        help="The data formats to emit, comma-separated. Valid values are 'parquet' and 'csv'. default='parquet,csv'",
-        type=str,
+        help="The data formats to emit, comma-separated. Default='parquet'",
+        default=EmitType.PARQUET,
+        type=EmitType,
+        choices=list(EmitType),
     )
     parser.add_argument(
         "--dryrun",
         help="Run the pipeline without actually executing any steps and inspect the configuration.",
         action="store_true",
     )
-    parser.add_argument("--nocache", help="Disable LLM cache.", action="store_true")
+    parser.add_argument(
+        "--nocache", help="Disable LLM cache.", action="store_true", default=False
+    )
     parser.add_argument(
         "--init",
         help="Create an initial configuration in the given path.",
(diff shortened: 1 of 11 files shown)


graphrag/index/progress/load_progress_reporter.py
```diff
@@ -19,12 +19,13 @@ def load_progress_reporter(reporter_type: str = "none") -> ProgressReporter:
...
-        return RichProgressReporter("GraphRAG Indexer ")
-    if reporter_type == "print":
-        return PrintProgressReporter("GraphRAG Indexer ")
-    if reporter_type == "none":
-        return NullProgressReporter()
-
-    msg = f"Invalid progress reporter type: {reporter_type}"
-    raise ValueError(msg)
+    match reporter_type:
+        case "rich":
```
andresmor-ms, line 23: You could use the ReporterType enum in here instead of strings

graphrag/prompt_tune/__main__.py
```diff
@@ -5,126 +5,116 @@
...
+        raise argparse.ArgumentTypeError(msg)
+    return path
+
+
+def dir_exist(path):
+    """Check for directory existence."""
+    if not Path(path).is_dir():
+        msg = f"Directory not found: {path}"
+        raise argparse.ArgumentTypeError(msg)
+    return path
```
andresmor-ms, line 29: You could move these functions into a utils to avoid duplication of code

graphrag/index/__main__.py
```diff
@@ -4,73 +4,123 @@
...
+from enum import Enum
+from pathlib import Path
 
 from .cli import index_cli
 
+
+class ReporterType(Enum):
+    """The type of reporter to use."""
+
+    RICH = "rich"
```
AlonsoGuevara, line 16: I've always wanted to replace our enum strings with auto(), would make the implementation cleaner.
WDYT? Not suggesting this as a refactor, but a conversation starter.

AlonsoGuevara: BTW, fully aware this is a cli PR :) I'm speaking in general hehe

graphrag/index/__main__.py
```diff
@@ -79,20 +129,20 @@
     args = parser.parse_args()
 
     if args.resume and args.update_index:
-        msg = "Cannot resume and update a run at the same time."
```
AlonsoGuevara, line 82: Interesting. Previously Ruff failed when there wasn't a dot. Guess it now changed :)

graphrag/index/api.py
```diff
@@ -22,12 +22,12 @@
 
 async def build_index(
     config: GraphRagConfig,
+    emit: list[str],
```
AlonsoGuevara, line 25: do you think it makes sense, at least from a conceptual standpoint, to have this param so high in the signature?

graphrag/index/__main__.py
```diff
@@ -4,73 +4,123 @@
...
+
+    PARQUET = "parquet"
+    CSV = "csv"
+
+    def __str__(self):
+        """Return the string representation of the enum value."""
+        return self.value
+
+
+def file_exist(path):
```
AlonsoGuevara, line 36: THis looks like it could go into the top level utils module

graphrag/query/__main__.py
```diff
@@ -22,69 +23,74 @@ def __str__(self):
         return self.value
 
 
+def file_exist(path):
```
AlonsoGuevara, line 26: yup, top level utils :)

graphrag/index/__main__.py
jgbradley1: Sounds like a good idea initially however I think we need the string representation for the CLI. I don't think `argparse` can display/handle enums very cleanly without an overloaded `__str__` or requiring more code.

For internal library operations, `auto` would be cleaner but then we're maintaining two different styles of enums which is worse in my opinion ☹️.

graphrag/prompt_tune/__main__.py
jgbradley1: Done. Moved to the top-level utils directory.

graphrag/index/__main__.py
jgbradley1: Done. Moved to the top-level utils directory.

graphrag/query/__main__.py
jgbradley1: Done. Moved to the top-level utils directory.

graphrag/index/progress/load_progress_reporter.py
jgbradley1: Done

graphrag/index/api.py
jgbradley1: I originally moved this due to removing the default value of `None`. It's been moved back though and the default value set to `parquet`. At the API layer, I think this makes more sense.

We don't want the defualt value to be defined as `None` at the API layer because that gets changed to `parquet` inside the library anyway. It's best to make this default behavior more transparent.

jgbradley1:
@AlonsoGuevara / @andresmor-ms I moved those functions into the `utils` directory. and incorporated more use of some existing Enum classes.

Let me know if there is any other feedback. Otherwise I think this is ready to merge.

AlonsoGuevara (approved):

PR #1140, opened 2024-09-16
Verb merge nre1

natoverse:
Begins the process of merging transformation verbs to simplify flos


graphrag/index/workflows/default_workflows.py
```diff
@@ -3,6 +3,9 @@
 
 """A package containing default workflows definitions."""
 
+# load and register all subflows
+from .v1.subflows import *  # noqa
```
andresmor-ms, line 7: If possible, lets avoid `* imports`, if you really want to import everything from this module you could do `import .v1.subflows as subflows` and use `subflows.join_text_units` or anything you need from that module

andresmor-ms: If I remember correctly,

you can use:

```python
from datashaper.verbs.engine import load_verbs

import .v1.subflows as subflow_verbs

load_verbs(subflow_verbs)

```

It's more verbose, but you avoid the * import. However, I now see that in other parts of graphrag we do use * imports so I guess it's fine here too :)

andresmor-ms (approved):

PR #1143, opened 2024-09-17
Collapse verbs: create_final_text_units

natoverse:
Next batch of collapsing - main addition here is the config loading so more complex workflows can be worked on


natoverse:
Update: since the join* workflows were only used by this one, I merged them into and removed all of their cruft.

graphrag/index/workflows/v1/subflows/create_final_text_units_pre_embedding.py
```diff
@@ -0,0 +1,178 @@
...
+                []
+                if not covariates_enabled
+                else [
+                    {
+                        "column": "covariate_ids",
+                        "operation": "any",
+                        "to": "covariate_ids",
+                    }
+                ]
+            ),
```
andresmor-ms, line 85: Not a fan of this, why not move out this list to a variable and then do:
```py
aggregations = [{ ... }]
if covariate_enabled:
    aggregations.append({...})
```

graphrag/index/workflows/v1/subflows/create_final_text_units_pre_embedding.py
```diff
@@ -0,0 +1,178 @@
...
+            right_on="id",
+            how="left",
+            suffixes=cast(Suffixes, ["_1", "_2"]),
+            indicator=True,
+        )
+    )
+
+
+def __clean_result(result: pd.DataFrame) -> pd.DataFrame:
+    result = cast(
```
andresmor-ms, line 170: This cast is not needed since the input is already a pd.DataFrame

andresmor-ms: Oh, nvm you are casting the pd.concat result

graphrag/index/workflows/v1/subflows/create_final_text_units_pre_embedding.py
```diff
@@ -0,0 +1,178 @@
...
+        pd.DataFrame,
+        pd.concat([
+            result[result["_merge"] == "both"],
+            result[result["_merge"] == "left_only"],
+            result[result["_merge"] == "right_only"],
```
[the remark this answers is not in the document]
andresmor-ms: From pandas docs:

> If True, adds a column to the output DataFrame called “_merge” with information on the source of each row. The column can be given a different name by providing a string argument. The column will have a Categorical type with the value of “left_only” for observations whose merge key only appears in the left DataFrame, “right_only” for observations whose merge key only appears in the right DataFrame, and “both” if the observation’s merge key is found in both DataFrames.

Which means we are concatenating all the possible values of `_merge` and then dropping the column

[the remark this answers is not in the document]
natoverse: Yes. I copied the util from DataShaper, which probably covers all strategies

graphrag/index/workflows/v1/subflows/create_final_text_units_pre_embedding.py
```diff
@@ -0,0 +1,178 @@
...
+    entity_joined = _join(selected, entity_join)
+    relationship_joined = _join(entity_joined, relationship_join)
+    final_joined = relationship_joined
+
+    if covariates_enabled:
+        final_covariates = others[2]
+        covariate_join = _covariates(final_covariates)
+        final_joined = _join(relationship_joined, covariate_join)
+
+    aggregated = aggregate_df(
```
AlonsoGuevara, line 47: Not sure if in scope for this PR, but surely we can optimize a lot if we don't use the aggregate verb but a different implementation. And it can be multiple one liners :)

graphrag/index/workflows/v1/subflows/create_final_text_units_pre_embedding.py
```diff
@@ -0,0 +1,178 @@
...
+
+def _join(left, right):
+    return __clean_result(
+        left.merge(
+            right,
+            left_on="id",
+            right_on="id",
+            how="left",
+            suffixes=cast(Suffixes, ["_1", "_2"]),
+            indicator=True,
```
AlonsoGuevara, line 164: I think we should only use this if we intend to do some distinction between rows origins, but since at the ends we are doing another concat we could just ignore this.

graphrag/index/workflows/v1/subflows/create_final_text_units_pre_embedding.py
natoverse: I will ignore for now so I can just focus on the collapsing, then we can do a separate sweep to cleanup and optimize a bunch of the pandas ops

AlonsoGuevara (approved):

andresmor-ms (approved):

PR #1150, opened 2024-09-17
Collapse final communities workflow

natoverse:


graphrag/index/workflows/v1/subflows/create_final_communities.py
```diff
@@ -0,0 +1,181 @@
...
+
+    graph_nodes = unpack_graph_df(table, callbacks, "clustered_graph", "nodes")
+    graph_edges = unpack_graph_df(table, callbacks, "clustered_graph", "edges")
+
+    source_clusters = graph_nodes.merge(
+        graph_edges,
+        left_on="label",
+        right_on="source",
+        how="inner",
+        suffixes=cast(Suffixes, ["_1", "_2"]),
```
AlonsoGuevara, line 46: No need to cast :)

graphrag/index/workflows/v1/subflows/create_final_communities.py
```diff
@@ -0,0 +1,181 @@
...
+        criteria=[
+            Criterion(
+                type=FilterCompareType.Column,
+                operator=StringComparisonOperator.Equals,
+                value="level_2",
+            ),
+        ],
+    )
+    filter_index = filter_df(table, args)
+    sub_idx = filter_index == True  # noqa: E712
```
AlonsoGuevara, line 134: Curious about this assignment

graphrag/index/workflows/v1/subflows/create_final_communities.py
```diff
@@ -0,0 +1,181 @@
...
+                ]
+            ],
+        )
+    )
+
+
+def _filter_clusters(table: pd.DataFrame) -> pd.DataFrame:
+    # level_1 is the left side of the join
+    # level_2 is the right side of the join
+    args = FilterArgs(
```
AlonsoGuevara, line 123: So the filter is, where level_1 == level_2 ?

AlonsoGuevara: Ah! I see, it is because of the suffixes haha. It confused me since level_1 and level_2 are also values.
Can we replace this verb invocation with a pandas one liner?

graphrag/index/workflows/v1/subflows/create_final_communities.py
natoverse: Pyright actually complained this time!

graphrag/index/workflows/v1/subflows/create_final_communities.py
natoverse: Yes :)

natoverse: We could actually probably just get rid of the suffixes and accept the pandas defaults (x/y), which may get rid of the weird casting issue too

graphrag/index/workflows/v1/subflows/create_final_communities.py
natoverse: Came from DataShaper - ask Andres ;)

graphrag/index/workflows/v1/subflows/create_final_communities.py
```diff
@@ -0,0 +1,181 @@
...
+            ),
+        ],
+    )
+    filter_index = filter_df(table, args)
+    sub_idx = filter_index == True  # noqa: E712
+    idx = filter_index[sub_idx].index  # type: ignore
+    return cast(pd.DataFrame, table[table.index.isin(idx)].reset_index(drop=True))
+
+
+def _create_title(table: pd.DataFrame) -> pd.DataFrame:
```
AlonsoGuevara, line 157: Please correct me if I'm wrong but pretty sure we can replace this and it aux functions by doing this:

```python
def _create_title(table: pd.DataFrame) -> pd.DataFrame:
    # Directly concatenate the "Community" string with the "id" column, handling any necessary type conversion
    table["title"] = "Community" + table["id"].astype(str)

    return table
   ```

File: graphrag/index/workflows/v1/subflows/create_final_communities.py [hunk 5]
```diff
@@ -0,0 +1,181 @@
...
+        ],
+    )
+    filter_index = filter_df(table, args)
+    sub_idx = filter_index == True  # noqa: E712
+    idx = filter_index[sub_idx].index  # type: ignore
+    return cast(pd.DataFrame, table[table.index.isin(idx)].reset_index(drop=True))
+
+
+# we've filtered on the left, now we filter on the right
+def _filter_secondary(table: pd.DataFrame) -> pd.DataFrame:
```
AlonsoGuevara, line 140: So, I think the simple version of this would look like:
```python
def _filter_secondary(table: pd.DataFrame) -> pd.DataFrame:
    # Directly filter rows where "level" column equals "level_1"
    return table[table["level"] == "level_1"].reset_index(drop=True)
   ```

File: graphrag/index/workflows/v1/subflows/create_final_communities.py [hunk 6]
```diff
@@ -0,0 +1,181 @@
...
+        suffixes=cast(Suffixes, ["_1", "_2"]),
+    )
+
+    filtered = _filter_secondary(joined)
+
+    titled = _create_title(filtered)
+
+    titled["raw_community"] = titled[
+        "id"
+    ]  # TODO: Rodrigo says "raw_community" is temporary
```
AlonsoGuevara, line 102: Since this was for the old streamlit implementation, we could validate if it is still needed.

natoverse: Let's just get rid of it - any old code that might have used it can use the `id` column

graphrag/index/workflows/v1/subflows/create_final_communities.py
AlonsoGuevara: So odd. This allows for a list of str, even a tuple of str. Ugh... Pyright...

graphrag/index/workflows/v1/subflows/create_final_communities.py
natoverse: That's great - a good example of how the flexible verb added a lot of cruft to a simple operation.

Slight correction - we are comparing the column values to each other, not to a literal:

`table[table["level"] == table["level_1"]].reset_index(drop=True)`

graphrag/index/workflows/v1/subflows/create_final_communities.py
natoverse: Got rid of all of it with the inline filter

AlonsoGuevara (approved):

PR #1152, opened 2024-09-18
[bugs fix] check conversation_history_context's value on community_context.py 

ZeyuTeng96:
## Description

For the following code (line 90 - 96), conversation_history_context is concatenated with community_context, but the case where conversation_history_context is empty("") has not been considered. When conversation_history_context is empty (""), concatenation should not be performed, as it would result in community_context or each element in community_context having an extra "\n\n".

```python
if isinstance(community_context, list):
     final_context = [
           f"{conversation_history_context}\n\n{context}"
           for context in community_context
      ]
else:
     final_context = f"{conversation_history_context}\n\n{community_context}"
```

## Proposed Changes

introducing a context_prefix to check the state of conversation_history_context, concatenation can be handled appropriately. When conversation_history_context is empty (""), the following code will use "" for concatenation. When conversation_history_context is not empty (""), the functionality will be similar to the previous code.

```python
context_prefix = f"{conversation_history_context}\n\n" if conversation_history_context else ""

if isinstance(community_context, list):
     final_context = [
           f"{context_prefix}{context}"
           for context in community_context
      ]
else:
     final_context = f"{context_prefix}{community_context}""
```

## Checklist

- [✔] I have tested these changes locally.
- [✔ ] I have reviewed the code changes.
- [✔ ] I have updated the documentation (if necessary).

## Additional Notes

N/A


AlonsoGuevara:
Merged in #1262

PR #1161, opened 2024-09-19
Fix example in documentation to ensure it runs correctly.

junho85:
## Description

The updated code initializes the async function properly, imports necessary modules, and ensures the pipeline execution runs as intended. This makes the example functional and aligns with the expected usage of the `run_pipeline` method.

## Checklist


jgbradley1:
We recently have been going through the docs and cleaning/removing a lot of content that is no longer relevant. We decided to remove this section of content entirely for the sake of simplicity and maintainability.

PR #1176, opened 2024-09-20
Collapse create base documents

natoverse:
- Collapses the workflow into one
- Required inverting some input file logic due to DataShaper's independent input injection
- Removes now-unused merge_overrides verb


graphrag/index/workflows/v1/subflows/create_base_documents.py
```diff
@@ -0,0 +1,85 @@
...
+        right_on="id",
+        how="right",
+    )
+    rejoined.rename(columns={"text": "raw_content"}, inplace=True)
+    rejoined["id"] = rejoined["id"].astype(str)
+
+    # attribute columns are converted to strings and then collapsed into a single json object
+    if document_attribute_columns is not None and len(document_attribute_columns) > 0:
+        for column in document_attribute_columns:
+            rejoined[column] = rejoined[column].astype(str)
```
andresmor-ms, line 72: I think that 
```py
rejoined = rejoined.astype({column: str for column in document_attribute_columns})
```
 would do the trick... the idea would be that the astype is executed just once on the dataframe instead of `len(document_attribute_columns)` times

graphrag/index/workflows/v1/subflows/create_base_documents.py
```diff
@@ -0,0 +1,85 @@
...
+        rejoined["attributes"] = rejoined[document_attribute_columns].apply(
+            lambda row: ({**row}),
+            axis=1,
+        )
+        rejoined.drop(columns=document_attribute_columns, inplace=True)
+        rejoined.reset_index()
```
andresmor-ms, line 78: you might want to do inplace=True here or do rejoined = rejoined.reset_index()

graphrag/index/workflows/v1/subflows/create_base_documents.py
```diff
@@ -0,0 +1,85 @@
...
+        source,
+        left_on="id",
+        right_on="id",
+        how="right",
+    )
+    rejoined.rename(columns={"text": "raw_content"}, inplace=True)
+    rejoined["id"] = rejoined["id"].astype(str)
+
+    # attribute columns are converted to strings and then collapsed into a single json object
+    if document_attribute_columns is not None and len(document_attribute_columns) > 0:
```
andresmor-ms, line 70: this condition could be replaced by `if document_attribute_columns:` (empty lists in python are False, and None is also False)

graphrag/index/workflows/v1/subflows/create_base_documents.py
```diff
@@ -0,0 +1,85 @@
...
+        for column in document_attribute_columns:
+            rejoined[column] = rejoined[column].astype(str)
+        rejoined["attributes"] = rejoined[document_attribute_columns].apply(
+            lambda row: ({**row}),
```
andresmor-ms, line 74: I think those parentheses are not required here

graphrag/index/workflows/v1/subflows/create_base_documents.py
```diff
@@ -0,0 +1,85 @@
...
+                "operation": "array_agg",
+                "to": "text_units",
+            }
+        ],
+    )
+
+    rejoined = docs_with_text_units.merge(
+        source,
+        left_on="id",
+        right_on="id",
```
andresmor-ms, line 63: you could use `on="id"` and remove the `left_on` and `right_on`

tests/verbs/util.py
```diff
@@ -65,17 +73,23 @@ def compare_outputs(
...
+        print("Expected:", cols)
+        print("Actual:", actual.columns)
         raise
 
+    for column in cols:
+        assert column in actual.columns
+        try:
+            # dtypes can differ since the test data is read from parquet and our workflow runs in memory
+            assert_series_equal(actual[column], expected[column], check_dtype=False)
+        except AssertionError:
```
andresmor-ms, line 86: Just curious, if you add `-vvv` when you execute the test, doesn't it automatically give you the expected vs actual?

andresmor-ms: Also you can do 

```py
assert condition, "Error message"
```

if you want a custom error message

graphrag/index/workflows/v1/subflows/create_base_documents.py
natoverse: This seems to leave the extra `index` column if I try it either of those ways

graphrag/index/workflows/v1/subflows/create_base_documents.py
natoverse: Checks complain about dict comprehension with this. It works fine from an equivalency perspective

tests/verbs/test_create_base_documents.py
```diff
@@ -0,0 +1,58 @@
...
+        {
+            "steps": steps,
+        },
+    )
+
+    # we should have dropped "title" and added "attributes"
+    # our test dataframe does not have attributes, so we'll assert without it
+    # and separately confirm it is in the output
+    compare_outputs(actual, expected, columns=["id", "text_units", "raw_content"])
+    assert len(actual.columns) == 4
```
AlonsoGuevara, line 57: To improve readability, we could do something like
```python
assert len(actual.columns) == 4, f"Expected 4 columns, but got {len(actual.columns)}"
```

Not for this PR tho, but, it is something we can take into account for refining our testing and debugging

AlonsoGuevara (approved):

tests/verbs/test_create_base_documents.py
natoverse: Great idea - lemme make a separate "test clarity pr" today

PR #1178, opened 2024-09-20
Collapse create base text units

natoverse:


graphrag/index/workflows/v1/subflows/create_base_text_units.py
```diff
@@ -0,0 +1,87 @@
...
+) -> VerbResult:
+    """All the steps to transform base text_units."""
+    table = cast(pd.DataFrame, input.get_input())
+
+    sort = table.sort_values(by=["id"], ascending=[True])
+
+    sort["text_with_ids"] = list(
+        zip(*[sort[col] for col in ["id", "text"]], strict=True)
+    )
+    sort.reset_index(drop=True)
```
andresmor-ms, line 40: also this has no effect on sort, when using reset_index() if you don't assign the result or add the inplace keyword, then reset_index call will reset the index but it won't store the result anywhere

graphrag/index/workflows/v1/subflows/create_base_text_units.py
```diff
@@ -0,0 +1,87 @@
...
+    chunked = genid_df(
+        chunked, to="chunk_id", method="md5_hash", hash=[chunk_column_name]
+    )
+
+    chunked[["document_ids", chunk_column_name, n_tokens_column_name]] = pd.DataFrame(
+        chunked[chunk_column_name].tolist(), index=chunked.index
+    )
+    chunked["id"] = chunked["chunk_id"]
+
+    filtered = chunked[chunked[chunk_column_name].notna()].reset_index(drop=True)
```
andresmor-ms, line 80: I think this is the same as chunked.dropna(subset=[chunk_column_name]).reset_index(drop=True) I like this version more because what you want to do is drop the na rows on a specific column

graphrag/index/verbs/genid.py
```diff
@@ -52,6 +52,19 @@ def genid(
...
+    output = genid_df(data, to, method, hash)
+
+    return TableContainer(table=output)
+
+
+def genid_df(
+    input: pd.DataFrame,
+    to: str,
+    method: str = "md5_hash",
+    hash: list[str] = [],  # noqa A002
```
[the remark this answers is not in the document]
andresmor-ms: I'd rewrite it this way:

```python
def genid_df(
    input: pd.DataFrame,
    to: str,
    method: str = "md5_hash",
    hash: list[str] | None = None,
):
    data = input
    
    match method:
        case "md5_hash":
            if not hash:
                msg = 'Must specify the "hash" columns to use md5_hash method'
                raise ValueError(msg)
            data[to] = data.apply(lambda row: gen_md5_hash(row, hash), axis=1)
        case "increment":
            data[to] = data.index + 1
        case _:
            msg = f"Unknown method {method}"
            raise ValueError(msg)
    
    return data

```

I changed the `if len(hash) == 0:` to `if not hash:` (this will check if hash is None and if len(hash) > 0) 

That way if the hash param is not provided and the method is md5_hash it will raise the ValueError.

 Also using match/case instead of if/elif/else...

graphrag/index/workflows/v1/subflows/create_base_documents.py
```diff
@@ -0,0 +1,85 @@
...
+    # attribute columns are converted to strings and then collapsed into a single json object
+    if document_attribute_columns is not None and len(document_attribute_columns) > 0:
+        for column in document_attribute_columns:
+            rejoined[column] = rejoined[column].astype(str)
+        rejoined["attributes"] = rejoined[document_attribute_columns].apply(
+            lambda row: ({**row}),
+            axis=1,
+        )
+        rejoined.drop(columns=document_attribute_columns, inplace=True)
+        rejoined.reset_index()
```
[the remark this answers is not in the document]
natoverse: This didn't work for me on other pr

graphrag/index/workflows/v1/subflows/create_base_text_units.py
natoverse: This seems to have downstream issues with a following verb util so I'm going to punt for now

AlonsoGuevara:
I thiiink this PR got also the changes from the collapse documents. Can we merge that one in and then update this one?

AlonsoGuevara (approved):

PR #1214, opened 2024-09-25
Reorganize python package structure

jgbradley1:
## Description
This PR is the first of what we expect to be several PRs that attempts to cleanup some of the package organization/design to allow for better flexibility. The general goal we are shooting for is to have a decoupled design made up of multiple components where each component can be swapped out for future solutions if the need arises. Some design choices are inspired by other popular OSS projects such as [mmdetection](https://github.com/open-mmlab/mmdetection).

## Proposed Changes

* api layer has been moved to a top-level `api` directory
* debugger configuration profiles have been defined to enable easier debugging with VSCode python extension
* all reporter code has been moved to a top-level `logging` directory with a more explicit factory design
* callbacks have been moved to a top-level `callbacks` directory with a more explicit factory design

## Checklist

## Notes
Some of the logging code should be cleaned up and consolidated in a future PR. Changes in this PR are mostly focused on high-level package organization first.


graphrag/logging/factories.py
```diff
@@ -1,18 +1,18 @@
...
 from .types import (
-    NullProgressReporter,
-    PrintProgressReporter,
     ProgressReporter,
     ReporterType,
 )
 
 
-def load_progress_reporter(
+def create_progress_reporter(
```
andresmor-ms, line 15: Maybe a silly question but, what's the difference between `logger` and `reporter` ?

graphrag/logging/__init__.py
```diff
@@ -0,0 +1,27 @@
+# Copyright (c) 2024 Microsoft Corporation.
+# Licensed under the MIT License
+
+"""Logging utilities and implementations."""
+
+from .console import ConsoleReporter  # noqa: I001
```
andresmor-ms, line 6: Why don't we sort the imports here?

graphrag/logging/factories.py
jgbradley1: That's a great question. The way we use it, I don't believe there is a difference. I personally would like to see a move toward a _logger_ naming convention throughout the codebase instead of _reporter_ as it aligns more with the python logging utility. And in line with the zen of python, explicit is better than implicit. I'm open to suggestions though!

Refactoring the code to use logger* names would make this particular PR too large (I tried it) so I instead chose to focus on just code organization for this PR as a first step.

I'd like to transform the factory functions for logging into a proper class instead of just functions to improve/enforce some of the organization and extensibility around using a factory design pattern

graphrag/logging/__init__.py
jgbradley1: Thanks for catching this. I previously encountered an error with circular imports that required these imports to be un-ordered but that appears to no longer be an issue. Once the factory design is fully implemented for certain top-level modules, the risk of circular imports will be even less.

andresmor-ms (dismissed):

graphrag/callbacks/global_search_callbacks.py
```diff
@@ -3,9 +3,10 @@
 
 """GlobalSearch LLM Callbacks."""
 
-from graphrag.query.llm.base import BaseLLMCallback
 from graphrag.query.structured_search.base import SearchResult
 
+from .llm_callbacks import BaseLLMCallback
```
andresmor-ms, line 8: can we use an absolute import here to match the one above it?

andresmor-ms: (not a fan of relative imports and mixing them) - but they sometimes resolve circular dependencies so if it's possible to be consistent it would be great :)

natoverse (approved):

PR #1219, opened 2024-09-25
Implement dynamic community selection for global search

bryanlimy:
## Description

Implement dynamic community selection to improve global search result. 

Currently, we select the community reports by specifying a specific community level, prior to the map-reduce operation, while fast, this operation could eliminate useful community reports (i.e. at lower levels) and/or include reports that are not informative to the query.

Here, we take advantage of the hierachical structure of the indexed dataset. We first ask the LLM to rate how relevant each level 0 community is with respect to the user query, we then traverse down the child node(s) if the current community report is deemed relevant. Ultimately reducing the number of community reports that get passed to the map-reduce operation in global search.

The updated [example_notebooks/global_search.ipynb](https://github.com/bryanlimy/graphrag/blob/dynamic-search/examples_notebooks/global_search.ipynb) show an example of such reduction in community reports with the same user query.

## Related Issues

## Proposed Changes

## Checklist

## Additional Notes


bryanlimy:
@microsoft-github-policy-service agree company="Microsoft"

graphrag/query/factories.py
```diff
@@ -114,7 +115,7 @@ def get_local_search_engine(
     """Create a local search engine based on data + configuration."""
     llm = get_llm(config)
     text_embedder = get_text_embedder(config)
-    token_encoder = tiktoken.get_encoding(config.encoding_model)
+    token_encoder = tiktoken.encoding_for_model(config.llm.model)
```
jgbradley1, line 118: Should we do this automatic lookup of encoding model via the LLM model name or rely on the user to provide the correct value?

This will error out for anyone not using OpenAI models.

Perhaps we compromise here and use the user-provided encoding model name as before. However we add a check to see if the LLM model name exists in the tiktoken library and if not, log a warning message to the user that the encoding model name is not recognized and we will fallback to the user-provided ending model name.

if the encoding model name IS in tiktoken and does not match the user-provided value, then should we error out or just log a warning?

graphrag/query/factories.py
```diff
@@ -160,16 +161,24 @@ def get_global_search_engine(
...
     reports: list[CommunityReport],
     entities: list[Entity],
+    communities: list[Community],
     response_type: str,
+    dynamic_selection: bool,
 ) -> BaseSearch:
     """Create a global search engine based on data + configuration."""
-    token_encoder = tiktoken.get_encoding(config.encoding_model)
+    llm = get_llm(config)
+    token_encoder = tiktoken.encoding_for_model(config.llm.model)
```
jgbradley1, line 170: Same comment as above regarding how we want to handle non-OpenAI models

graphrag/index/verbs/graph/report/restore_community_hierarchy.py
```diff
@@ -35,7 +35,7 @@ def restore_community_hierarchy(
         name = row[name_column]
         community = row[community_column]
 
-        if community_levels.get(level) is None:
+        if level not in community_levels:
             community_levels[level] = {}
         community_levels[level][community] = name
```
AlonsoGuevara, line 40: Won't this error out if the level was just added as an empty dict?

graphrag/query/__main__.py
```diff
@@ -11,6 +11,7 @@
 from .cli import run_global_search, run_local_search
 
 INVALID_METHOD_ERROR = "Invalid method"
+DYNAMIC_SELECTION_ERROR = "Dynamic community selection only supports global search"
```
AlonsoGuevara, line 14: If that's the case, should we even select a method when doing dynamic selection?

.semversioner/next-release/major-20240923234633073836.json
```diff
@@ -0,0 +1,4 @@
+{
+  "type": "major",
```
AlonsoGuevara, line 2: minor or patch, not major

jgbradley1: Recommend patch as it’s an update to an existing query strategy.

bryanlimy:
Re-open the issue under https://github.com/microsoft/graphrag/pull/1364

PR #1240, opened 2024-10-01
Reorganize flows

natoverse:
Starts the process of moving core flows to a non-DataShaper structure. They are still wrapped in verbs for execution at the moment, but all of the subflows will now be extractable without any verb wrappings.

### DIFF
--- a/graphrag/index/flows/create_base_entity_graph.py
+++ b/graphrag/index/flows/create_base_entity_graph.py
@@ -1,0 +1,76 @@
+# Copyright (c) 2024 Microsoft Corporation.
+# Licensed under the MIT License
+
+"""All the steps to create the base entity graph."""
+
+from typing import Any, cast
+
+import pandas as pd
+from datashaper import (
+    VerbCallbacks,
+)
+from datashaper.table_store.types import VerbResult, create_verb_result
+
+from graphrag.index.storage import PipelineStorage
+from graphrag.index.verbs.graph.clustering.cluster_graph import cluster_graph_df
+from graphrag.index.verbs.graph.embed.embed_graph import embed_graph_df
+from graphrag.index.verbs.snapshot_rows import snapshot_rows_df
+
+
+async def create_base_entity_graph(
+    source: pd.DataFrame,
+    callbacks: VerbCallbacks,
+    storage: PipelineStorage,
+    clustering_config: dict[str, Any],
+    embedding_config: dict[str, Any],
+    graphml_snapshot_enabled: bool = False,
+    embed_graph_enabled: bool = False,
+) -> pd.DataFrame:
+    """All the steps to create the base entity graph."""
+
+    clustering_strategy = clustering_config.get("strategy", {"type": "leiden"})
+
+    clustered = cluster_graph_df(
+        source,
+        callbacks,
+        column="entity_graph",
+        strategy=clustering_strategy,
+        to="clustered_graph",
+        level_to="level",
+    )
+
+    if graphml_snapshot_enabled:
+        await snapshot_rows_df(
+            clustered,
+            column="clustered_graph",
+            base_name="clustered_graph",
+            storage=storage,
+            formats=[{"format": "text", "extension": "graphml"}],
+        )
+
+    embedding_strategy = embedding_config.get("strategy")
+    if embed_graph_enabled and embedding_strategy:
+        clustered = await embed_graph_df(
+            clustered,
+            callbacks,
+            column="clustered_graph",
+            strategy=embedding_strategy,
+            to="embeddings",
+        )
+
+    # take second snapshot after embedding
+    # todo: this could be skipped if embedding isn't performed, other wise it is a copy of the regular graph?
+    if graphml_snapshot_enabled:
+        await snapshot_rows_df(
+            clustered,
+            column="entity_graph",
+            base_name="embedded_graph",
+            storage=storage,
+            formats=[{"format": "text", "extension": "graphml"}],
+        )
+
+    final_columns = ["level", "clustered_graph"]
+    if embed_graph_enabled:
+        final_columns.append("embeddings")
+
+    return clustered[final_columns]
--- a/graphrag/index/flows/create_base_entity_graph.py
+++ b/graphrag/index/flows/create_base_entity_graph.py
@@ -9,7 +9,6 @@ import pandas as pd
 from datashaper import (
     VerbCallbacks,
 )
-from datashaper.table_store.types import VerbResult, create_verb_result
 
 from graphrag.index.storage import PipelineStorage
 from graphrag.index.verbs.graph.clustering.cluster_graph import cluster_graph_df
@@ -27,7 +26,6 @@ async def create_base_entity_graph(
     embed_graph_enabled: bool = False,
 ) -> pd.DataFrame:
     """All the steps to create the base entity graph."""
-
     clustering_strategy = clustering_config.get("strategy", {"type": "leiden"})
 
     clustered = cluster_graph_df(
@@ -73,4 +71,4 @@ async def create_base_entity_graph(
     if embed_graph_enabled:
         final_columns.append("embeddings")
 
-    return clustered[final_columns]
+    return cast(pd.DataFrame, clustered[final_columns])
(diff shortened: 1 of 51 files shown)


graphrag/index/flows/create_final_relationships.py
```diff
@@ -0,0 +1,85 @@
...
+            column="description",
+            strategy=text_embed["strategy"],
+            embedding_name="relationship_description",
+        )
+
+    pruned_edges = filtered.drop(columns=["level"])
+
+    filtered_nodes = cast(
+        pd.DataFrame,
+        nodes[nodes["level"] == 0].reset_index(drop=True)[["title", "degree"]],
```
andresmor-ms, line 51: this looks weird to me haha, I'd split this in two lines:

```py
filtered_nodes = cast(pd.DataFrame, nodes[nodes["level"] == 0])
filtered_nodes = filtered_nodes[["title", "degree"]].reset_index(drop=True)
```

graphrag/index/flows/create_final_relationships.py
```diff
@@ -0,0 +1,85 @@
...
+        node_name_column="title",
+        node_degree_column="degree",
+        edge_source_column="source",
+        edge_target_column="target",
+    )
+
+    edge_combined_degree["human_readable_id"] = edge_combined_degree[
+        "human_readable_id"
+    ].astype(str)
+    edge_combined_degree["text_unit_ids"] = _to_array(
```
andresmor-ms, line 67: I think this can be as simple as:

edge_combined_degree["text_unit_ids"] = edge_combined_degree["text_unit_ids"] .str.split(",")

graphrag/index/flows/create_base_entity_graph.py
```diff
@@ -0,0 +1,74 @@
...
+        clustered = await embed_graph_df(
+            clustered,
+            callbacks,
+            column="clustered_graph",
+            strategy=embedding_strategy,
+            to="embeddings",
+        )
+
+    # take second snapshot after embedding
+    # todo: this could be skipped if embedding isn't performed, other wise it is a copy of the regular graph?
```
AlonsoGuevara, line 60: Is this todo new? Or legacy?

graphrag/index/flows/create_final_text_units.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+def _entities(df: pd.DataFrame) -> pd.DataFrame:
+    selected = df.loc[:, ["id", "text_unit_ids"]]
+    unrolled = selected.explode(["text_unit_ids"]).reset_index(drop=True)
+
+    return (
+        unrolled.groupby("text_unit_ids", sort=False)
+        .agg(entity_ids=("id", "unique"))
+        .reset_index()
```
andresmor-ms, line 85: we probably want to add drop=True on this one?

graphrag/index/flows/create_final_text_units.py
```diff
@@ -0,0 +1,120 @@
...
+        .rename(columns={"text_unit_id": "id"})
+    )
+
+
+def _join(left, right):
+    return left.merge(
+        right,
+        left_on="id",
+        right_on="id",
```
andresmor-ms, line 117: can remove these in favor of `on="id"`

graphrag/index/flows/create_final_text_units.py
```diff
@@ -0,0 +1,120 @@
...
+                *(
+                    []
+                    if (not text_embed or is_using_vector_store)
+                    else ["text_embedding"]
+                ),
+                "n_tokens",
+                "document_ids",
+                "entity_ids",
+                "relationship_ids",
+                *([] if final_covariates is None else ["covariate_ids"]),
```
andresmor-ms, line 72: I wonder if we can change these optional things with:
`"covariate_ids" if final_covariates else *[]` Not sure if this is valid syntax haha

graphrag/index/flows/create_final_relationships.py
AlonsoGuevara: Soooo, I wouldn't sweat too much about pandas optimizations as I'll be sweeping through this as soon as Nathan gives me green light

graphrag/index/flows/create_base_entity_graph.py
natoverse: The step is legacy, the todo I added so we would remember to review it. Basically it looks like that snapshot would be identical to the first snapshot if embedding isn't performed, so it seems like it should be condition on embedding happening.

graphrag/index/flows/create_final_community_reports.py
```diff
@@ -0,0 +1,167 @@
...
+from uuid import uuid4
+
+import pandas as pd
+from datashaper import (
+    AsyncType,
+    VerbCallbacks,
+)
+
+from graphrag.index.cache import PipelineCache
+from graphrag.index.graph.extractors.community_reports.schemas import (
```
AlonsoGuevara, line 15: I think I did a note to self in a previous PR, but, will add it here as well: Simplify this import

graphrag/index/flows/create_final_text_units.py
natoverse: In general we always want drop=True, right? I can't think of a use case where we want to retain a`index` column

graphrag/index/flows/create_final_community_reports.py
```diff
@@ -0,0 +1,167 @@
...
+        community_hierarchy,
+        callbacks,
+        cache,
+        strategy,
+        async_mode=async_mode,
+        num_threads=num_threads,
+    )
+
+    community_reports["id"] = community_reports["community"].apply(
+        lambda _x: str(uuid4())
```
AlonsoGuevara, line 86: Wonder if we could replace this apply with someting like:
`community_reports["id"] = pd.util.hash_pandas_object(community_reports.index).astype(str)`

These ids are just random, right?

graphrag/index/flows/create_final_text_units.py
natoverse: Oh, nevermind - in this case we do want to retain it, because of the aggregation operation. The index column appears to become `text_unit_ids`, which we rename to `id` and then use for the join

graphrag/index/flows/create_final_community_reports.py
natoverse: I do not know what uniqueness guarantees pandas makes with that function versus uuid

graphrag/index/flows/create_final_community_reports.py
```diff
@@ -0,0 +1,167 @@
...
+            cache,
+            column="title",
+            strategy=title_text_embed["strategy"],
+            embedding_name="community_report_title",
+        )
+
+    return community_reports
+
+
+def _prep_nodes(input: pd.DataFrame) -> pd.DataFrame:
```
AlonsoGuevara, line 125: Note to self: Remove this apply

graphrag/index/flows/create_final_text_units.py
natoverse: I'm not a fan of these inline array destructures, but they kind of litter the place. Let's take a "syntax cleanliness" pass later

graphrag/index/flows/create_final_community_reports.py
```diff
@@ -0,0 +1,167 @@
...
+            NODE_NAME: x[NODE_NAME],
+            NODE_DESCRIPTION: x[NODE_DESCRIPTION],
+            NODE_DEGREE: x[NODE_DEGREE],
+        },
+        axis=1,
+    )
+    return input
+
+
+def _prep_edges(input: pd.DataFrame) -> pd.DataFrame:
```
AlonsoGuevara, line 140: Idem. Rule of thumb: Only use apply when there's no other way

graphrag/index/flows/create_final_covariates.py
```diff
@@ -0,0 +1,69 @@
...
+        callbacks,
+        column,
+        covariate_type,
+        strategy,
+        async_mode,
+        entity_types,
+        num_threads,
+    )
+
+    covariates["id"] = covariates["covariate_type"].apply(lambda _x: str(uuid4()))
```
AlonsoGuevara, line 45: Note to self: Remove this

graphrag/index/flows/create_final_community_reports.py
natoverse: Andres also seemed to think these could be simplified with a to_dict, but it didn't work on first try so needs a little more investigation

AlonsoGuevara (approved):

PR #1242, opened 2024-10-02
Migrate embedding operations

natoverse:
Moves two verbs (text and graph embedding) into the new DataShaper verb-less "operations" folder. The intent is to move all of the actively used verb functions in this manner, and delete all of the verb wrappings and unused verbs. Also in this migration, I will attempt to introduce a little more functional naming consistency, as well as move off of kwargs for better usage clarity.


graphrag/index/flows/create_final_text_units.py
```diff
@@ -42,16 +42,16 @@ async def create_final_text_units(
     aggregated = final_joined.groupby("id", sort=False).agg("first").reset_index()
 
     is_using_vector_store = False
-    if text_embed:
-        aggregated["text_embedding"] = await text_embed_df(
+    if text_text_embed:
```
AlonsoGuevara, line 45: wonder why the extra text_?

natoverse: Because the naming convention on each of these is {field}_text_embed

AlonsoGuevara (approved):

graphrag/index/flows/create_final_text_units.py
AlonsoGuevara: Sound like moon moon haha, but approved

PR #1247, opened 2024-10-03
Update final nodes output

AlonsoGuevara:
## Description

Update create_final_nodes

## Related Issues

## Proposed Changes

## Checklist

## Additional Notes

### DIFF
--- a/graphrag/index/update/dataframes.py
+++ b/graphrag/index/update/dataframes.py
@@ -5,6 +5,11 @@
 
 import os
 from dataclasses import dataclass
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from collections.abc import Callable
+
 
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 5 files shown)


graphrag/index/update/dataframes.py
```diff
@@ -309,3 +328,206 @@ def _update_and_merge_text_units(
...
+
+    Parameters
+    ----------
+    new_delta_nodes_df : pd.DataFrame
+        The new delta nodes.
+    merged_relationships_df : pd.DataFrame
+        The merged relationships.
+    old_nodes : pd.DataFrame
+        The old nodes.
+    community_count_threshold : int, optional
```
ha2trinh, line 479: should we have some params to define the % of existing relationships vs new relationships (e.g. if a new node has 2 old neighbors and 10 new neighbors then we consider putting it in a new community, rather than an old community?

AlonsoGuevara: At the time I'm doing simple majority, and not even fine tuning it haha. I  totally agree that should be percentage and not an int. to be independent on the dataset size.
My plan is to further refine this value once we have community reports generated.
But, will address the change of converting to a percentage right now

andresmor-ms (approved):

graphrag/index/update/dataframes.py
```diff
@@ -309,3 +328,213 @@ def _update_and_merge_text_units(
...
+        old_nodes.loc[:, "title"].isin(node_relationships["source"])
+        | old_nodes.loc[:, "title"].isin(node_relationships["target"])
+    ]
+
+    # Merge with new_delta_nodes_df to get the level and community info
+    related_communities = related_communities.merge(
+        new_delta_nodes_df[["level", "title"]], on=["level", "title"], how="outer"
+    )
+
+    # Count the communities for each (level, title) pair
```
ha2trinh, line 495: maybe we remove all these logics for calculating community_counts too?

PR #1248, opened 2024-10-03
Migrate helper verbs

natoverse:
Moves most of the simpler verbs to non-DataShaper functional structure. Flattens things in a few places. Deletes many unused verbs,


natoverse:
Note that this PR has expanded to include _all_ of the primary verbs, not just the simplest ones. The only one remaining is the community report generation, which will wait until incremental index is complete to avoid conflicts. This PR also does not tackle flattening the tests that reference these verbs.

AlonsoGuevara (approved):

PR #1267, opened 2024-10-10
Migration script for text embeddings

gaudyb:
## Description

Migration script to transform previous generated indexing files into a separate files per column embedding and original source file without embedding columns.

## Proposed Changes

extract_text_embeddings.ipynb script


migration_scripts/extract_text_embeddings.ipynb
```diff
@@ -0,0 +1,156 @@
+{
```
andresmor-ms, line 1: If this is intended to be executed multiple times, could we make it a standalone .py file that acts as a CLI tool with arguments? this could even go into the graphrag source code.

PR #1270, opened 2024-10-11
Fix support for ollama and other llms

liuda-666:
## Description

This great program currently supports OpenAI services, which seems to limit its development. In actual enterprise production activities, we usually localize the deployment of some open source models and services because it can effectively protect privacy data. Based on the above requirements, I have changed the part of the large model calling to ollama. I hope you can consider my opinions and needs. Or add new branch sections, and in addition, I have also fixed some minor issues that I have discovered in practice.

### DIFF
--- a/graphrag/llm/openai/openai_chat_llm.py
+++ b/graphrag/llm/openai/openai_chat_llm.py
@@ -6,7 +6,7 @@
 import logging
 
 from typing_extensions import Unpack
-
+from openai import AsyncOpenAI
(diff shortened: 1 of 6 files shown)


liuda-666:
microsoft-github-policy-service agree [company="8lab"]

JoedNgangmeni:
@liuda-666 Please! They're waiting on you i think

natoverse:
See [new docs](https://microsoft.github.io/graphrag/config/models/) on injecting custom model providers

Issue #1275, opened 2024-10-13
[Bug]: performance improvements in map_query_to_entities()


mmaitre314:
### Do you need to file an issue?

- [x] I have searched the existing issues and this bug is not already filed.
- [x] My model is hosted on OpenAI or Azure. If not, please look at the "model providers" issue and don't file a new one here.
- [x] I believe this is a legitimate bug, not just a question. If this is a question, please use the Discussions area.

### Describe the bug

Doing some profiling to find opportunities to speed up queries, local search spends a good amount of time in
`get_entity_by_key()`: 7s out of 20s for 49K entities, with another 9s spent waiting for GPT-4o to generate the response. That method makes an O(N) scan on the entity list when at least in the default case where `embedding_vectorstore_key == EntityVectorStoreKey.ID` it could do a O(1) lookup in the entity dictionary. In a quick test, replacing `matched = get_entity_by_key(...)` by `matched = all_entities_dict.get(result.document.id)` effectively made those 7s go away. Also, in the general case of O(N) full scans, since `value` is constant in `get_entity_by_key()`, calls to
`isinstance()`, `is_valid_uuid()`, and `replace()` could be moved out of the loop to reduce the hot spot.

### Steps to reproduce

Run local search with 50K entities.

### Expected Behavior

Most of the query time is spent on generating the AI summary.

### GraphRAG Config Used

```python
token_encoder = tiktoken.get_encoding("cl100k_base")

local_search = LocalSearch(
    llm = ChatOpenAI(
        azure_ad_token_provider=azure_ad_token_provider,
        model="gpt-4o",
        api_base="https://<snip>.openai.azure.com",
        api_version="2024-02-15-preview",
        api_type=OpenaiApiType.AzureOpenAI,
        max_retries=1,
    ),
    context_builder = LocalSearchMixedContext(
        community_reports = reports,
        text_units = text_units,
        entities = entities,
        relationships = relationships,
        covariates = None,
        entity_text_embeddings = description_embedding_store,
        embedding_vectorstore_key = EntityVectorStoreKey.ID,
        text_embedder = OpenAIEmbedding(
            azure_ad_token_provider=azure_ad_token_provider,
            api_base="https://<snip>.openai.azure.com",
            api_type=OpenaiApiType.AzureOpenAI,
            api_version="2024-02-15-preview",
            model="text-embedding-ada-002",
            deployment_name="text-embedding-ada-002",
            max_retries=1,
        ),
        token_encoder = token_encoder,
[...]

PR #1276, opened 2024-10-13
Perf optimizations in map_query_to_entities()

mmaitre314:
## Description

Perf optimizations in `map_query_to_entities()`.

Result of a perf benchmark using PyTest on a list with 1M entities (corresponding roughly to calling `map_query_to_entities()` with 50K entities and the defaults `k = 10` and `oversample_scaler = 2`):
- Original `get_entity_by_key()`: 2.15s
- Optimized `get_entity_by_key()`: 0.10s
- Lookup `get_entity_by_id()`: 0.00s (below PyTest's duration truncation)

## Related Issues

#1275

## Proposed Changes

- In the default case where `embedding_vectorstore_key == EntityVectorStoreKey.ID`, use the fact that entities are already stored in a dictionary to perform an O(1) lookup instead of an O(N) scan. The lookup is implemented in a new method called `get_entity_by_id()`.
- In the general case, optimize `get_entity_by_key()` by moving `isinstance()`, `is_valid_uuid()`, `replace()` out of the loop and calling `getattr()` once instead of twice.

## Checklist


mmaitre314:
@microsoft-github-policy-service agree company="Microsoft"

AlonsoGuevara (approved):

PR #1273, opened 2024-10-11
DRIFT Search

j2whiting:
## Description

This PR adds a dynamic, graph reasoning module for performing GraphRAG queries.

## Proposed Changes

* Create configuration and defaults for DRIFT
* Update BaseSearch to accept generic context builder. This is done to allow for a different signature for the DRIFT context builder while still allowing it to be a child of BaseSearch
* Create Misc. Utilities for running DRIFT query, storing state, serializing, warm starting, etc.

## Checklist

- [x] I have tested these changes locally.
- [x] I have reviewed the code changes.
- [ ] I have updated the documentation (if necessary). **Detailed docs are coming shortly**
- [ ] I have added appropriate unit tests (if applicable). **Let me know if you think we should add some tests here, it's a pretty big PR and touches a decent chunk of the repo. Perhaps the search APIs are good candidates for unit tests?**

## Additional Notes

A few more experiments are in the pipeline to extend DRIFT functionality. The QueryState and DRIFTAction objects were built to accommodate experimenting with different forms of reasoning chains. More to come soon..

PR #1285, opened 2024-10-15
DRIFT Search 

AlonsoGuevara:
## Description

All kudos to @j2whiting on #1273 

This PR adds a dynamic, graph reasoning module for performing GraphRAG queries.

## Proposed Changes

* Create configuration and defaults for DRIFT
* Update BaseSearch to accept generic context builder. This is done to allow for a different signature for the DRIFT context builder while still allowing it to be a child of BaseSearch
* Create Misc. Utilities for running DRIFT query, storing state, serializing, warm starting, etc.

## Checklist

- [ ] I have updated the documentation (if necessary). **Detailed docs are coming shortly**
- [ ] I have added appropriate unit tests (if applicable). **Let me know if you think we should add some tests here, it's a pretty big PR and touches a decent chunk of the repo. Perhaps the search APIs are good candidates for unit tests?**

## Additional Notes

A few more experiments are in the pipeline to extend DRIFT functionality. The QueryState and DRIFTAction objects were built to accommodate experimenting with different forms of reasoning chains. More to come soon..


graphrag/model/community_report.py
```diff
@@ -62,3 +62,30 @@ def from_dict(
             full_content_embedding=d.get(full_content_embedding_key),
             attributes=d.get(attributes_key),
         )
+
+    def to_dict(
```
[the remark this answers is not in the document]
AlonsoGuevara: Yup, they can and I'll follow your advise :)
I wanted to keep the custom key behavior the class has but... Pretty sure only defaults are used

j2whiting (approved):
LGTM 

Thanks @AlonsoGuevara

graphrag/query/structured_search/drift_search/search.py
```diff
@@ -0,0 +1,290 @@
...
+        ------
+        RuntimeError: If no intermediate answers or follow-up queries are found in the primer response.
+        """
+        response = search_results.response
+        if isinstance(response, list) and isinstance(response[0], dict):
+            intermediate_answers = [
+                i["intermediate_answer"] for i in response if "intermediate_answer" in i
+            ]
+
+            if len(intermediate_answers) == 0:
```
andresmor-ms, line 123: Could be replaced with: `if not intermediate_answers`

graphrag/query/structured_search/drift_search/state.py
```diff
@@ -0,0 +1,141 @@
...
+        action: DriftAction,
+        follow_ups: list[DriftAction] | list[str],
+        weight: float = 1.0,
+    ):
+        """Add all follow-up actions and links them to the given action."""
+        if len(follow_ups) == 0:
+            log.warning("No follow-up actions for action: %s", action.query)
+
+        for follow_up in follow_ups:
+            if isinstance(follow_up, str):
```
andresmor-ms, line 45: According to the method signature it should be a list of DriftActions, not str

andresmor-ms: nvm, didn't see the | list[str]. but still, if the type is the first one, list[DriftAction] and it throws a warning, why allow it?

AlonsoGuevara: Yup, fixing that actually

andresmor-ms (approved):

andresmor-ms (approved):

PR #1291, opened 2024-10-17
Empty workflow returns

natoverse:
Adds a simple empty check for workflow dataframe (None is not allowed per DataShaper API), and avoids writing result to storage. This allows us to create side-effects workflows with no written return value.

### DIFF
--- a/graphrag/index/run/workflow.py
+++ b/graphrag/index/run/workflow.py
@@ -19,7 +19,7 @@ from graphrag.callbacks.progress_workflow_callbacks import ProgressWorkflowCallb
 from graphrag.index.context import PipelineRunContext
 from graphrag.index.emit.table_emitter import TableEmitter
 from graphrag.index.run.profiling import _write_workflow_stats
-from graphrag.index.storage.typing import PipelineStorage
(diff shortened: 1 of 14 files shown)


graphrag/index/run/workflow.py
```diff
@@ -48,8 +48,13 @@ async def _emit_workflow_output(
 ) -> pd.DataFrame:
     """Emit the workflow output."""
     output = cast(pd.DataFrame, workflow.output())
-    for emitter in emitters:
-        await emitter.emit(workflow.name, output)
+    rows, cols = output.shape
+    # only write the final output if it has content
+    # this is expressly designed to allow us to create
+    # workflows with side effects that don't produce a formal output to save
+    if rows > 0 and cols > 0:
```
jgbradley1, line 55: What about `if not output.empty`? The [docs](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.empty.html) suggest this better accounts for the edge case where there may be columns but no rows in a dataframe.

natoverse: This is why we have better pandas users than myself review!

jgbradley1 (approved):

PR #1296, opened 2024-10-18
New workflow to generate embeddings in a single workflow

gaudyb:
## Description

New workflow to separate text embedding generation in a new single workflow to generate new parquet files with just id/embedding columns.

## Proposed Changes

New workflow called: create_final_embeddings.
Remove all embeddings configuration in all other workflows.
Standardize text embed configurations for all embeddings.
New standard for parquet embeddings files: `<original_table_name>_<source_column>_embeddings.parquet` with just id/embedding columns.


graphrag/config/enums.py
```diff
@@ -86,6 +86,7 @@ class TextEmbeddingTarget(str, Enum):
 
     all = "all"
     required = "required"
+    none = "none"
```
AlonsoGuevara, line 89: Is this valid? We would still need entity description_embeddings for local search to work

graphrag/index/flows/create_final_embeddings.py
```diff
@@ -0,0 +1,211 @@
...
+    final_entities: pd.DataFrame,
+    final_community_reports: pd.DataFrame,
+    callbacks: VerbCallbacks,
+    cache: PipelineCache,
+    storage: PipelineStorage,
+    embedded_fields: set[str],
+    base_text_embed: dict | None = None,
+) -> pd.DataFrame:
+    """All the steps to generate all embeddings."""
+    documents_embeddings = final_documents[["id", "raw_content"]].copy()
```
AlonsoGuevara, line 45: Do we need to copy these values? Can we just mantain references?

graphrag/index/flows/create_final_embeddings.py
```diff
@@ -0,0 +1,211 @@
...
+    )
+
+    community_reports_embeddings = final_community_reports[
+        ["id", "full_content", "summary", "title"]
+    ].copy()
+
+    if base_text_embed:
+        log.info("Creating embeddings")
+        # Embed raw_content in documents table
+        if document_raw_content_embedding in embedded_fields and isinstance(
```
AlonsoGuevara, line 61: This repeated code could be extracted to a function

graphrag/config/enums.py
gaudyb: @natoverse wanted to implement this version with no embeddings at all for a specific case

graphrag/index/flows/create_final_embeddings.py
```diff
@@ -0,0 +1,211 @@
...
+                base_text_embed=base_text_embed,
+                message="Text units embeddings (text)",
+            )
+
+        # Embed name in entities table
+        if entity_name_embedding in embedded_fields and isinstance(
+            entities_embeddings, pd.DataFrame
+        ):
+            await run_embeddings(
+                data=entities_embeddings.copy(),
```
AlonsoGuevara, line 110: This was already a copy, and now this calls copy again. II would even challenge the need to copy at all

graphrag/index/flows/create_final_embeddings.py
```diff
@@ -0,0 +1,211 @@
...
+            data,
+            callbacks,
+            cache,
+            column=column_to_embed,
+            strategy=base_text_embed["strategy"],
+        )
+
+        data = cast(pd.DataFrame, data[["id", "embedding"]])
+
+    await snapshot(
```
AlonsoGuevara, line 206: I don't think running embedding should also store the file. These are separate concerns.

graphrag/index/run/profiling.py
```diff
@@ -69,7 +69,3 @@ async def _write_workflow_stats(
         await _save_profiler_stats(
             storage, workflow.name, workflow_result.memory_profile
         )
-
-    log.debug(
```
AlonsoGuevara, line 73: Why are we removing this from profiling?

gaudyb: I already checked this with @natoverse yesterday, because now the pipeline supports empty dataframes.

graphrag/index/workflows/v1/subflows/create_final_text_units.py
```diff
@@ -43,14 +38,11 @@ async def create_final_text_units(
     if final_covariates:
         final_covariates = cast(pd.DataFrame, final_covariates.table)
 
-    output = await create_final_text_units_flow(
+    output = create_final_text_units_flow(
```
AlonsoGuevara, line 41: Is this not async anymore?

graphrag/index/flows/create_final_embeddings.py
gaudyb: remember that now the embeddings are not in memory at all, the full process ends at disk write so I believe it should be in the same function.

graphrag/index/flows/create_final_embeddings.py
AlonsoGuevara: And loop through some params, to avoid that much repetition

gaudyb: the repetitive call is just a function call, the only repetitive code is the if statement itself and we want to maintain this until we find a more generic process, right @natoverse ?

graphrag/index/workflows/v1/subflows/create_final_text_units.py
gaudyb: No, the only async method inside it was the text_embed one so no async anymore.

graphrag/index/flows/create_final_embeddings.py
AlonsoGuevara: The disagree. The purpose of the function is to run embeddings and store it on a column. Why mis it with I/O operations in the same context?

graphrag/index/flows/create_final_embeddings.py
```diff
@@ -0,0 +1,211 @@
...
+        log.info("%s", message)
+        data["embedding"] = await embed_text(
+            data,
+            callbacks,
+            cache,
+            column=column_to_embed,
+            strategy=base_text_embed["strategy"],
+        )
+
+        data = cast(pd.DataFrame, data[["id", "embedding"]])
```
AlonsoGuevara, line 204: I think this cast isn't necessary

graphrag/index/flows/create_final_embeddings.py
```diff
@@ -0,0 +1,211 @@
...
+                filename="create_final_documents_raw_content_embeddings",
+                callbacks=callbacks,
+                cache=cache,
+                storage=storage,
+                base_text_embed=base_text_embed,
+                message="Documents embeddings (raw content)",
+            )
+
+        # Embed description in relationships table
+        if relationship_description_embedding in embedded_fields and isinstance(
```
AlonsoGuevara, line 76: I think we don't need to run the instance check. These params are already data frames, and there's slicing done above. If not data frames, at this point code would've already crashed.

graphrag/index/flows/create_final_embeddings.py
gaudyb: it was requested by ruff in the format check

graphrag/index/flows/create_final_embeddings.py
gaudyb: yes I agree but the ruff check requested the validation to not throw errors.

graphrag/index/flows/create_final_embeddings.py
gaudyb: because this column won't be saved in memory at all, the embeddings is just an I/O operation now.

graphrag/index/flows/create_final_embeddings.py
AlonsoGuevara: you can use methods than ensure data frame retrieval. For example, instead of df["val] do df.loc[:, "val"]

graphrag/index/flows/create_final_embeddings.py
AlonsoGuevara: Same as below, if we use slicing that ensures df retrieval we can avoid type checks and casting

graphrag/index/flows/create_final_embeddings.py
gaudyb: If I don't copy them the workflow execution fails because python throws an error that the dataframe should be copied in order to manipulate it

graphrag/index/flows/create_final_embeddings.py
gaudyb: same answer than above

graphrag/index/flows/create_final_embeddings.py
AlonsoGuevara: What I mean is in this function we want to return data, and then call snapshot in another place after calling run_embeddings. So run_embeddings calculates embedding fields without side effect like saving to disk which should be the concern of this method or wouldn't be expected.

graphrag/index/flows/create_final_embeddings.py
AlonsoGuevara: Same as below, if we use slicing that ensures df retrieval we can avoid type checks and casting

graphrag/index/flows/create_final_embeddings.py
gaudyb: @natoverse what do you think?

graphrag/index/flows/create_final_embeddings.py
```diff
@@ -0,0 +1,194 @@
...
+    filename: str,
+    callbacks: VerbCallbacks,
+    cache: PipelineCache,
+    storage: PipelineStorage,
+    base_text_embed: dict | None = None,
+    message: str | None = None,
+) -> None:
+    """All the steps to generate single embedding."""
+    if base_text_embed:
+        log.info("%s", message)
```
AlonsoGuevara, line 178: Oh, check for msg. What will happen if it is null? Would still print something or just ignore it?

natoverse: I'm not really a fan of having this message arg anyway. It only exists for info logging, which is a bit weird to put in an "api". How about just logging using the filename, which contains all of the relevant info?

graphrag/index/flows/create_final_embeddings.py
natoverse: I agree with both points. I think Alonso is right that it feels weird to bundle these two concerns into a single function. However, because the entire point of this workflow is to run + save the embeddings, this function more serves to reduce duplicate code than anything, so it's not bothering me much. A couple thoughts:

- It would help to rename it something like "run_and_snapshot_embeddings" if we keep them bundled
- It also might help to put `_` at the front to indicate it is an internal function for the module
- Can we measure Gaudy's point? If the function call indirection unloads the temporary dataframe from the call stack and reduces peak memory, this definitely seems worth bundling into one function.
- To the last point: up-front in the flow is the creation of a bunch of selected dataframes, these could each be put after the if statement for their associated embeddings to co-locate the logic, and maybe save memory

graphrag/index/run/profiling.py
natoverse: @AlonsoGuevara this call was just a debug, which now fails because it assumes a dataframe with at least one row. The call to the profiler happens above anyway. We can put in a "null check" if you think it's really helpful

graphrag/index/workflows/v1/create_final_entities.py
```diff
@@ -17,26 +20,12 @@ def build_steps(
...
     """
-    base_text_embed = config.get("text_embed", {})
-    entity_name_embed_config = config.get("entity_name_embed", base_text_embed)
-    entity_name_description_embed_config = config.get(
-        "entity_name_description_embed", base_text_embed
-    )
-
-    skip_name_embedding = config.get("skip_name_embedding", False)
-    skip_description_embedding = config.get("skip_description_embedding", False)
+    log.info(config)
```
natoverse, line 23: Probably don't need this log anymore

graphrag/index/workflows/v1/create_final_relationships.py
```diff
@@ -18,19 +22,11 @@ def build_steps(
     * `workflow:create_base_entity_graph`
     * `workflow:create_final_nodes`
     """
-    base_text_embed = config.get("text_embed", {})
-    relationship_description_embed_config = config.get(
-        "relationship_description_embed", base_text_embed
-    )
-    skip_description_embedding = config.get("skip_description_embedding", False)
+    log.info(config)
```
natoverse, line 25: Same. I would remove any logging that is put in for dev/debug and not useful for regular runs of the system

natoverse:
Please remember to rename the new workflow to remove "create_final" so it is more clear that it doesn't have a direct output.

natoverse:
Please add a test under test_verbs that confirms all selected embeddings tables are created. You can test this by providing a storage instance in the test, and then confirming that all requested tables exist. You can also read any of the tables and assert that they have mock embeddings values.

graphrag/index/flows/create_final_embeddings.py
```diff
@@ -0,0 +1,194 @@
...
+    )
+
+    community_reports_embeddings = final_community_reports.loc[
+        :, ["id", "full_content", "summary", "title"]
+    ]
+
+    if base_text_embed:
+        log.info("Creating embeddings")
+        # Embed raw_content in documents table
+        if document_raw_content_embedding in embedded_fields:
```
andresmor-ms, line 60: Could we replace all these if's for a loop?

graphrag/index/flows/create_final_embeddings.py
```diff
@@ -0,0 +1,194 @@
...
+    text_units_embeddings = final_text_units.loc[:, ["id", "text"]]
+
+    entities_embeddings = final_entities.loc[:, ["id", "name", "description"]]
+    entities_embeddings["name_description"] = (
+        entities_embeddings["name"] + ":" + entities_embeddings["description"]
+    )
+
+    community_reports_embeddings = final_community_reports.loc[
+        :, ["id", "full_content", "summary", "title"]
+    ]
```
andresmor-ms, line 55: What do you think about doing this preprocessing at the verb level so that this function only takes what is needed to work?

graphrag/config/enums.py
AlonsoGuevara: We should just update the init file, so it add none to the options then

graphrag/index/flows/create_final_embeddings.py
andresmor-ms:
```python

embedding_param_map = {
    document_raw_content_embedding: {
        "data": documents_embeddings,
        "column_to_embed"="raw_content"
        "filename": "create_final_documents_raw_content_embeddings",
        "message": "Documents embeddings (raw content)"
    },
    # and so on for every embedding
}

for field in embedded_fields:
    await run_embeddings(
        callbacks=callbacks,
        cache=cache,
        storage=storage,
        base_text_embed,
        **embedding_param_map[field]
    )
...

```

Bonus points if the embedding_param_map is a constant outside of this function and then on this function you create a map for the data parameter to make it cleaner.

AlonsoGuevara: Exactly my thoughts, there's a lot of repetition than can be refactor within a loop

AlonsoGuevara (changes requested):
Overall comment, I don't like that we just remove the embeddings smoke tests. Can you please add tests to check for these outputs?

graphrag/index/flows/create_final_embeddings.py
natoverse: Yeah, Gaudy and I talked about a loop being the eventual form

.vscode/launch.json
```diff
@@ -19,8 +19,8 @@
 			"module": "poetry",
 			"args": [
 				"poe", "query",
-				"--root", "<path_to_ragtest_root_demo>",
-				"--method", "global",
+				"--root", "/Users/gaudy-microsoft/Desktop/christmas_carol-new-version",
```
AlonsoGuevara, line 22: This is pointing to something in your local machine

graphrag/vector_stores/lancedb.py
```diff
@@ -16,6 +15,7 @@
     VectorStoreDocument,
     VectorStoreSearchResult,
 )
+import lancedb as lancedb
```
jgbradley1, line 18: Minor comment...I don't understand this import. Wouldn't `import lancedb` do the same thing?

.vscode/launch.json
jgbradley1: I reverted this back to the old setting.

jgbradley1 (approved):

graphrag/index/flows/generate_text_embeddings.py
```diff
@@ -0,0 +1,153 @@
...
+    log.info("Creating embeddings")
+    for field in embedded_fields:
+        await _run_and_snapshot_embeddings(
+            callbacks=callbacks,
+            cache=cache,
+            storage=storage,
+            **embedding_param_map[field],
+        )
+
+    return pd.DataFrame()
```
andresmor-ms, line 124: If this function always returns an empty dataframe why not remove it? looks weird to execute a function to get an empty dataframe... I'm guessing then that _run_and_snapshot_embeddings has a side effect? I'd like to make it clear that this function does not return anything useful by removing the return statement, is that possible?

graphrag/vector_stores/lancedb.py
gaudyb: That's the import issue we have been facing and reporting in the chat

graphrag/index/flows/generate_text_embeddings.py
gaudyb: No, according to the return structure workflow is expecting, that's what Nathan explained to me.

graphrag/vector_stores/lancedb.py
AlonsoGuevara: But what Josh means is the as lancedb

graphrag/index/flows/generate_text_embeddings.py
AlonsoGuevara: Andrés has a great point tho. This should be in the verb, not in the core logic. It couples this to datashaper

andresmor-ms: IMHO it makes no sense to have this return here, what is expected is for the verb to return a dataframe, I'd move this pd.DataFrame() directly to the create_verb_output function in the verb to avoid polluting the new code. @natoverse what do you think?

graphrag/index/flows/generate_text_embeddings.py
```diff
@@ -0,0 +1,153 @@
...
+            data,
+            callbacks,
+            cache,
+            column=column_to_embed,
+            strategy=base_text_embed["strategy"],
+        )
+
+        data = data.loc[:, ["id", "embedding"]]
+
+    await snapshot(
```
andresmor-ms, line 148: Is it expected that if we don't run the embed_text function we still save the file?

graphrag/index/flows/generate_text_embeddings.py
gaudyb: @natoverse we talked about this in the past, could you explain to @andresmor-ms please

graphrag/index/flows/generate_text_embeddings.py
```diff
@@ -0,0 +1,153 @@
...
+            callbacks=callbacks,
+            cache=cache,
+            storage=storage,
+            **embedding_param_map[field],
+        )
+
+    return pd.DataFrame()
+
+
+async def _run_and_snapshot_embeddings(
```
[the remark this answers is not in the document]
gaudyb: At least from my point of view snapshot can't be optional but @natoverse could you please give your opinion in this

graphrag/config/defaults.py
```diff
@@ -82,6 +82,7 @@
 SNAPSHOTS_GRAPHML = False
 SNAPSHOTS_RAW_ENTITIES = False
 SNAPSHOTS_TOP_LEVEL_NODES = False
+SNAPSHOTS_EMBEDDINGS = True
```
natoverse, line 85: This should be False by default

graphrag/index/create_pipeline_config.py
```diff
@@ -332,37 +259,103 @@ def _graph_workflows(
...
+        PipelineWorkflowReference(
+            name=generate_text_embeddings,
+            config={
+                "document_raw_content_embed": _get_embedding_settings(
+                    settings.embeddings,
+                    "document_raw_content",
+                    {
+                        "title_column": "raw_content",
+                        "collection_name": "final_documents_raw_content_embedding",
+                        "snapshot": settings.snapshots.embeddings,
```
natoverse, line 313: Move these to a single top-level config just like the graphml one (it can be called embeddings_snapshot to keep the same pattern). That way it does not need to be repeated, nor does it need to be dug out of the deep config later in the flow.

graphrag/index/flows/generate_text_embeddings.py
```diff
@@ -0,0 +1,163 @@
...
+            data,
+            callbacks,
+            cache,
+            column=column_to_embed,
+            strategy=base_text_embed["strategy"],
+        )
+
+        data = data.loc[:, ["id", "embedding"]]
+
+        if base_text_embed["strategy"]["vector_store"]["snapshot"] is True:
```
natoverse, line 157: See above about `embeddings_snapshot` param, which will just be a top-level bool

graphrag/index/workflows/v1/create_final_entities.py
natoverse, line 23: Remove log

graphrag/index/workflows/v1/create_final_relationships.py
natoverse, line 25: Remove log

natoverse (changes requested):

natoverse (dismissed):

AlonsoGuevara (approved):

graphrag/index/update/entities.py
xxll88, line 1: "Column(s) ['description_embedding'] do not exist"

graphrag/index/operations/embed_text/embed_text.py
```diff
@@ -142,24 +146,24 @@ async def _text_embed_with_vector_store(
...
-    title_column: str = vector_store_config.get("title_column", "title")
-    id_column: str = vector_store_config.get("id_column", "id")
+
     overwrite: bool = vector_store_config.get("overwrite", True)
 
-    if column not in input.columns:
+    if embed_column not in input.columns:
+        msg = f"Column {embed_column} not found in input dataframe with columns {input.columns}"
+        raise ValueError(msg)
+    title = title_column or embed_column
```
prasodani, line 155: While trying to index, embed_text fails due to this change.
`title` is re-initialised in the while loop below leading to a key error. Using some other variable name will fix it.

Issue #741, opened 2024-07-26
Incremental indexing (adding new content)


natoverse:
### Do you need to file an issue?

- [ ] I have searched the existing issues and this feature is not already filed.
- [ ] My model is hosted on OpenAI or Azure. If not, please look at the "model providers" issue and don't file a new one here.
- [ ] I believe this is a legitimate feature request, not just a question. If this is a question, please use the Discussions area.

### Is your feature request related to a problem? Please describe.

A number of users have asked how to add new content to an existing index without needing to re-run the entire process. This is a feature we are planning, and are in the design stages now to ensure we have an efficient approach.

As it stands, new content can be added to a GraphRAG index without requiring a complete re-index. This is because we rely heavily on a cache to avoid repeating the same calls to the model API. There are several stages within the pipeline for which this is very efficient - namely those stages that are atomic and do not have upstream dependencies. For example, if you add new documents to a folder, we will not re-chunk existing documents or perform entity and relationship extraction on the existing content; we will simply fetch the processed content from the cache and pass it on. The new documents _will_ be processed and new entities and relationships extracted. Downstream of this, the graph construction process will need to recreate the graph to include the new nodes and edges, and communities will be recomputed - resulting in re-summarization, etc. You can get a sense of this process and what downstream steps may be re-processed by looking at the [data flow diagram](https://microsoft.github.io/graphrag/posts/index/1-default_dataflow/).

### Describe the solution you'd like

An ideal solution would be to add a new command to GraphRAG such as `update` that can be run against new data and augment an existing index. Considerations here include things such as evaluating the new entities to determine if they can be added to an existing community, and when those communities have been altered enough to constitute a "drift" that needs recomputing. We could also perform analysis to determine which communities have been edited, such that we ignore summarization on those that haven't changed.

### Additional context

We also need to consider the types of analysis incremental ingest can enable beyond just "updates". For example, daily ingest of news with thoughtful graph construction/annotation could allow for delta analysis such that questions like "what happened with person _x_ in the last 24 hours" or "catch me up on the news themes this week".

natoverse:
This is a popular request, so I'm going to pin it and route other issues here.

natoverse:
Related: removing existing content, e.g., #585

KylinMountain:
can we split index into graph build and community summary? Lots of fans ask me if we can modify the knowledge graph in manual?

KylinMountain:
If you modify the Ilm params in settings.yaml, all of cache will be invalid.

[...]

PR #1318, opened 2024-10-24
Add Incremental Indexing v1

AlonsoGuevara:
## Description

Add Incremental Indexing support.

## Usage
Drop new files into the already existing input folder and update settings.yaml

In settings.yaml

```yaml
storage:
  type: file
  base_dir: "path/to/existing/artifacts"

update_index_storage:
  type: file
  base_dir: "path/to/create/new/index"
  
```

## Related Issues

#741 

## Checklist

## Additional Notes


graphrag/config/create_graphrag_config.py
```diff
@@ -375,6 +375,25 @@ def hydrate_parallelization_params(
...
+                    connection_string=reader.str(Fragment.conn_string),
+                    storage_account_blob_url=reader.str(
+                        Fragment.storage_account_blob_url
+                    ),
+                    container_name=reader.str(Fragment.container_name),
+                    base_dir=reader.str(Fragment.base_dir)
+                    or defs.UPDATE_STORAGE_BASE_DIR,
+                )
+            else:
+                update_index_storage_model = None
```
andresmor-ms, line 396: you could remove this else by moving the `update_index_storage_model = None` above the if :), could even go outside the with statement.

graphrag/index/flows/create_final_communities.py
```diff
@@ -61,6 +63,12 @@ def create_final_communities(
 
     filtered["title"] = "Community " + filtered["id"].astype(str)
 
+    # Add period timestamp to the community reports
+    filtered["period"] = datetime.now(timezone.utc).date().isoformat()
+
+    # Add size of the community
+    filtered["size"] = filtered.loc[:, "text_unit_ids"].apply(lambda x: len(x))
```
andresmor-ms, line 70: I think you can change it to `.apply(len)` no need for the lambda

AlonsoGuevara: I'll submit another PR optimizing all this calculations. This change was to attack some warnings being output

graphrag/config/create_graphrag_config.py
AlonsoGuevara: I had it like that but the issue was that the interpreter was assuming this to always be set. So it greyed out some of my none checks and was driving me nuts.
I can try again by setting explicit typings.

andresmor-ms (dismissed):

andresmor-ms (approved):

shamalgithub:
Thank you for this feature update ! 

Based on the instructions , i belive all one has to do is to add a new file to the 'input' folder and update the settings.yaml file with the new output file name.  
image

Looking at the issue , the proposed fix suggest that a graphrag.append command would be added to work with increamental indexing but that doesnot seem to  be present. 
image

I might be missing something here. Can someone kindly point out the correct way to do incremental indexing ?

joycerlz:
I also have the same question about the graphrag.append command, is it a valid command yet? Or we just add the file and update settings.yaml, no need to run additional commands? Any response/discussion would be appreciated, thanks!

PR #1341, opened 2024-10-31
Artifact cleanup

natoverse:
Cleans up a large number of duplicated or unused fields in the output artifacts, and renames a few things for consistency.


graphrag/index/update/entities.py
```diff
@@ -62,10 +63,9 @@ def _group_and_resolve_entities(
             "id": "first",
             "type": "first",
             "human_readable_id": "first",
-            "graph_embedding": "first",
             "description": lambda x: list(x.astype(str)),  # Ensure str
             # Concatenate nd.array into a single list
-            "text_unit_ids": lambda x: ",".join(str(i) for j in x.tolist() for i in j),
+            "text_unit_ids": lambda x: list(itertools.chain(*x.tolist())),
```
AlonsoGuevara, line 68: Nice :)

graphrag/utils/embeddings.py
```diff
@@ -0,0 +1,25 @@
...
+    Create a collection name for the embedding store.
+
+    Within any given vector store, we can have multiple sets of embeddings organized into projects.
+    The `container` param is used for this partitioning, and is added as a prefix to the collection name for differentiation.
+
+    The embedding name is fixed, with the available list defined in graphrag.index.config.embeddings
+
+    Note that we use dot notation in our names, but many vector stores do not support this - so we convert to dashes.
+    """
+    if validate and embedding_name not in all_embeddings:
```
AlonsoGuevara, line 22: Really like this util! We should use it also on the api methods (please dismiss if we already do)

graphrag/api/query.py
```diff
@@ -469,7 +476,7 @@ def _patch_vector_store(
             from graphrag.vector_stores.lancedb import LanceDBVectorStore
 
             community_reports = with_reports
-            collection_name = (
+            container_name = (
                 config.embeddings.vector_store.get("container_name", "default")
```
jgbradley1, line 480:
```suggestion
                config.embeddings.vector_store["container_name"]
```

I recommend to not add default values to fallback on in the python code as we plan to switch over to a better config system that will validate/enforce the existence of `container_name` in the config. It will be better and easier to debug if the python code just breaks when `container_name` is not present in the dict.

jgbradley1:
Now that the column names have been cleaned up quite a bit, is this a good PR to add docstrings to each workflow step to document the expected schema of the input/output dataframes?

As an alternative approach to docstrings, what do you think about using pydantic models to reference and validate the schema of each dataframe?

graphrag/index/flows/create_base_text_units.py
```diff
@@ -65,21 +63,18 @@ async def create_base_text_units(
...
-        chunked[chunk_column_name].tolist(), index=chunked.index
-    )
-    chunked["id"] = chunked["chunk_id"]
+    # rename for downstream consumption
+    chunked.rename(columns={"chunk": "text"}, inplace=True)
 
-    output = cast(
-        pd.DataFrame, chunked[chunked[chunk_column_name].notna()].reset_index(drop=True)
-    )
+    output = cast(pd.DataFrame, chunked[chunked["text"].notna()].reset_index(drop=True))
```
AlonsoGuevara, line 77: We can remove this cast by doing ```chunked = chunked.loc[chunked["text"].notna()].reset_index(drop=True)```

AlonsoGuevara (approved):

j2whiting (approved):

graphrag/api/query.py
natoverse: In this case I think I was just aiming for back-compat. Looking up the function a bit I see that we are creating this with defaults though, so that seems sensible. Once we enforce the new model we can take out the default creation entirely.

graphrag/index/flows/create_base_text_units.py
natoverse: I feel like we're finding a bunch of this. Thanks to pandas warnings, I think we've been doing it "wrong" by not using `loc`.

natoverse: This didn't quite work so I'm going to skip for now and add a v1 todo to remove as many casts as possible

natoverse:
> Now that the column names have been cleaned up quite a bit, is this a good PR to add docstrings to each workflow step to document the expected schema of the input/output dataframes?
> 
> As an alternative approach to docstrings, what do you think about using pydantic models to reference and validate the schema of each dataframe?

I would much prefer Pydantic models, as I've noticed some of those docstrings were out of date anyway. We can add that at any time though I think. FWIW, I also added a docs page for the parquet schemas to help answer user questions.

jgbradley1 (approved):

AlonsoGuevara (approved):

PR #1403, opened 2024-11-14
Fix non-default emitters

AlonsoGuevara:
## Description

Fix non-default emitters (Piepline assumed parquet always)

## Related Issues

## Proposed Changes

## Checklist

## Additional Notes

### DIFF
--- a/.semversioner/next-release/patch-20241114001527332372.json
+++ b/.semversioner/next-release/patch-20241114001527332372.json
@@ -1,0 +1,4 @@
+{
+  "type": "patch",
+  "description": "Fix issue when using emitters that are not the default parque output"
+}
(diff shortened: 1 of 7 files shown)


jgbradley1:
I tested and confirmed this addresses the bug on the indexing side. It breaks on the query side because the code [here](https://github.com/microsoft/graphrag/blob/main/graphrag/utils/storage.py#L44) assumes a parquet format is being loaded. I'm not sure if the query issue needs to be addressed or not though to enable the cosmosdb support.

AlonsoGuevara:
> I tested and confirmed this addresses the bug on the indexing side. It breaks on the query side because the code [here](https://github.com/microsoft/graphrag/blob/main/graphrag/utils/storage.py#L44) assumes a parquet format is being loaded. I'm not sure if the query issue needs to be addressed or not though to support the cosmosdb support.

Ah! Thanks for testing query. I'll address that on this same PR, great catch!!

AlonsoGuevara:
Closing this PR, as we discussed, this opened the possibility of more issues than fixes, so, .parquet will now be always included as output and emitters like json will be additional.

PR #1404, opened 2024-11-14
Move prompts

natoverse:
Moves the query and indexing prompts into a central location at root so they're easier to find. This also exports query prompts during `graphrag init` and loads them up from disk just like the indexing prompts, so end users can tune their queries easier.

A secondary add on this is putting DRIFT search config into the official config system, since it was not there but was necessary to configure the prompt on disk.

This PR does _not_ move or export the prompts for prompt_tune. That's a fair bit more work and much less likely to need user adjusting (the point is that it is auto...), so we'll consider that at a later time.


graphrag/index/init_content.py
```diff
@@ -177,6 +181,28 @@
   # map_max_tokens: {defs.GLOBAL_SEARCH_MAP_MAX_TOKENS}
   # reduce_max_tokens: {defs.GLOBAL_SEARCH_REDUCE_MAX_TOKENS}
   # concurrency: {defs.GLOBAL_SEARCH_CONCURRENCY}
+
+drift_search:
```
AlonsoGuevara, line 185: Most of these values have already been fine tuned, so not super sure if we want to expose them (given how big the config is now). Maybe something for discussion.

AlonsoGuevara (approved):

graphrag/index/init_content.py
natoverse: Will discuss with broader config convo

PR #1406, opened 2024-11-14
fix: correct path to images

SinghCoder:
## Description

The PR just fixes the path to the images referenced in markdown files.

## Proposed Changes

Just the image paths changed from `../img` to `./img`

## Checklist

- [X] I have updated the documentation (if necessary).


natoverse:
These paths are actually correct - the markdown pages render as a "folder" in the URL hierarchy (e.g.,: https://microsoft.github.io/graphrag/prompt_tuning/auto_prompt_tuning/), so you need the extra ../. This is a bit confusing during development because it seems one level too low.

PR #1411, opened 2024-11-15
First cut at config cleanup

natoverse:
Reorganizes the emitted `graphrag init` settings.yml to remove many of the lesser-used options so it is less confusing. Added much better documentation for this so the file can be stripped down.


docs/config/yaml.md
```diff
@@ -97,49 +79,79 @@ This is the base LLM configuration section. Other steps may override this config
...
-## reporting
+### update_index_storage
 
-### Fields
+#### Fields
+
+- `type` **file|memory|blob** - The storage type to use. Default=`file`
+- `connection_string` **str** - (blob only) The Azure Storage connection string.
+- `container_name` **str** - (blob only) The Azure Storage container name.
+- `base_dir` **str** - The base directory to write reports to, relative to the root.
```
AlonsoGuevara, line 139: To write outputs to

docs/prompt_tuning/manual_prompt_tuning.md
```diff
@@ -6,49 +6,85 @@ We provide a means for you to do this by allowing you to specify a custom prompt
...
+- **{response_type}** - Describe how the response should look (reducer only). We default to "multiple paragraphs".
+- **{context_data}** - The data tables from GraphRAG's index.
+
+### Drift Search
+
+[Prompt Source](http://github.com/microsoft/graphrag/blob/main/graphrag/prompts/query/drift_search_system_prompt.py)
+
+#### Tokens
+
+- **{response_type}** - Describe how the response should look. We default to "multiple paragraphs".
```
AlonsoGuevara, line 87: Right now Drift does not support response type, at least until I add the reduce prompt

graphrag/config/init_content.py
```diff
@@ -0,0 +1,141 @@
...
+
+parallelization:
+  stagger: {defs.PARALLELIZATION_STAGGER}
+  # num_threads: {defs.PARALLELIZATION_NUM_THREADS}
+
+async_mode: {defs.ASYNC_MODE.value} # or asyncio
+
+embeddings:
+  async_mode: {defs.ASYNC_MODE.value} # or asyncio
+  vector_store:{defs.VECTOR_STORE}
```
AlonsoGuevara, line 36: Do we document somewhere how to set up for AAIS?

natoverse: It is on the yaml page now

docs/prompt_tuning/manual_prompt_tuning.md
natoverse: The token is in the prompt though, so if someone is manually tuning they can write in what they want here.

graphrag/config/init_content.py
natoverse: The bullets don't nest well though, so it isn't clear. Checking...

AlonsoGuevara (approved):

PR #1452, opened 2024-11-27
Miscellaneous code cleanup procedures

jgbradley1:
## Description

This PR contains several miscellaneous code cleanup changes:

* move `graphrag.index.storage` -> `graphrag.storage`
* removed multiple emitter options - graphrag will only support exporting files in the parquet format when `storage.type=file`
* updated some of the terminology used in the code/docs/comments from *emitter* to *exporter* for better clarity. Open to reverting back to the old terminology if there are strong opinions against *exporter*.
* renamed all factory related files to `factory.py` for better consistency across the codebase
* refactored code under `graphrag.index.utils` to remove unnecessary "shortcut" functions that were simply calling factory methods. Other parts of the library should call factory methods directly to reduce complexity and increase the amount of shared code.

## Checklist


jgbradley1:
For better alignment, `graphrag.index.cache` should be moved up a level to `graphrag.cache` as well. I could do that in another PR or this one if we want to move it around.

jgbradley1:
I went ahead and moved `graphrag.index.cache` up a level to `graphrag.cache` to be in better alignment with the design we're shooting for.

natoverse (approved):

PR #1454, opened 2024-11-29
Update index API + a notebook that provides a general API overview

jgbradley1:
## Description

In this PR:
* an optional callbacks parameter is added to the index API for cleaner integration with the solution accelerator and other users.
* removed a hardcoded folder path name that was causing an empty directory to get created
* add a notebook that provides a demonstration of the Index + Query API

## Checklist


natoverse (dismissed):
Only suggestion would be to rename "api-overview" to "api_overview" for consistency with the others

jgbradley1:
> Only suggestion would be to rename "api-overview" to "api_overview" for consistency with the others

Good catch. I renamed the file.

AlonsoGuevara (approved):

PR #1456, opened 2024-11-30
Refactor Create Final Community reports to simplify code

AlonsoGuevara:
## Description

Clean up code and refactor methods.
Stabilize cache entries by sorting the report context every time it is created

## Related Issues

## Proposed Changes

## Checklist

## Additional Notes


graphrag/index/flows/create_final_community_reports.py
```diff
@@ -83,9 +84,7 @@ async def create_final_community_reports(
 
     community_reports["community"] = community_reports["community"].astype(int)
     community_reports["human_readable_id"] = community_reports["community"]
-    community_reports["id"] = community_reports["community"].apply(
-        lambda _x: str(uuid4())
-    )
+    community_reports["id"] = [str(uuid4()) for _ in range(len(community_reports))]
```
andresmor-ms, line 87: You could use `uuid4().hex` to get the value as string instead of `str()`

it is slightly faster:

```
%%timeit
[str(uuid4()) for _ in range(10000)]
``` 
61.8 ms ± 5.91 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

```
%%timeit
[uuid4().hex for _ in range(10000)]
``` 
50.3 ms ± 3.89 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

graphrag/index/graph/extractors/community_reports/sort_context.py
```diff
@@ -154,3 +154,81 @@ def _get_context_string(
...
+        ).drop_duplicates()
+        claims = pd.DataFrame(
+            group[claim_details_column].dropna().tolist()
+        ).drop_duplicates()
+        nodes = pd.DataFrame(
+            group[node_details_column].dropna().tolist()
+        ).drop_duplicates()
+
+        # Sort dataframes for consistent output
+        if not nodes.empty:
```
andresmor-ms, line 182: what happens if this is empty and we sort it? does it fails? could we just sort and remove the `if not empty` statements?

graphrag/index/graph/extractors/community_reports/sort_context.py
```diff
@@ -154,3 +154,81 @@ def _get_context_string(
...
+                by=[schemas.EDGE_DEGREE, schemas.EDGE_ID],
+                ascending=[False, True],
+                inplace=True,
+            )
+        if not claims.empty:
+            claims.sort_values(by=schemas.CLAIM_ID, ascending=True, inplace=True)
+
+        # Generate context string
+        contexts = []
+        if not nodes.empty and len(nodes) > 0:
```
andresmor-ms, line 199: is it possible for `not nodes.empty` to be True and `len(nodes)` = 0?

AlonsoGuevara: yup, empty checks for shape, so either having columns or rows works, and, in this case, we end up with columns. Maybe just checking len will be enoug

AlonsoGuevara: Dunno, let me check

natoverse (approved):

andresmor-ms (approved):

PR #1464, opened 2024-12-03
Graph collapse

natoverse:
Refactors the graph creation to finally eliminate serialized graphml.

- creates transient base nodes, edges, and communities
- moves graph embedding to create_final_nodes where it is actually used with umap
- removes lots of no-longer-applicable config


graphrag/index/flows/create_base_entity_graph.py
```diff
@@ -55,89 +52,121 @@ async def create_base_entity_graph(
...
+
+def _prep_nodes(entities, summaries, graph) -> pd.DataFrame:
+    degrees_df = _compute_degree(graph)
+    entities.drop(columns=["description"], inplace=True)
+    nodes = entities.merge(summaries, on="name", how="left")
+    nodes = nodes.merge(degrees_df, on="name")
+    nodes = nodes.drop_duplicates(subset="name")
+    nodes = nodes.loc[nodes["name"].notna()]
+    nodes.reset_index(inplace=True)
+    nodes.rename(columns={"name": "title", "source_id": "text_unit_ids"}, inplace=True)
```
andresmor-ms, line 142: just fyi, you can chain all of these operations

```python
nodes = entities.merge(...).merge(...).drop_duplicates(...).etc
```

graphrag/index/flows/create_base_entity_graph.py
```diff
@@ -55,89 +52,121 @@ async def create_base_entity_graph(
...
+    )
+    base_communities = base_communities.explode("title")
+    base_communities["community"] = base_communities["community"].astype(int)
+    return base_communities
+
+
+def _compute_degree(graph: nx.Graph) -> pd.DataFrame:
+    degrees = []
+    for node, degree in graph.degree:  # type: ignore
+        degrees.append({"name": node, "degree": int(degree)})
```
andresmor-ms, line 171: here you could use list-comprehension

```
return pd.DataFrame([{"name": node, "degree": int(degree)} for node, degree in graph.degree])
```

graphrag/index/flows/create_final_communities.py
```diff
@@ -4,82 +4,72 @@
...
+        entity_ids.groupby("community").agg(entity_ids=("id", list)).reset_index()
     )
 
-    target_clusters = graph_nodes.merge(
-        graph_edges, left_on="label", right_on="target", how="inner"
+    # aggregate relationships ids for each community
+    # these are limited to only those where the source and target are in the same community
+    max_level = base_communities["level"].max()
+    all_grouped = pd.DataFrame(
+        columns=cast(Any, ["community", "level", "relationship_ids", "text_unit_ids"])
```
andresmor-ms, line 29: What error were you getting here? I'm curious about the cast(Any, ...)  did you consider using `# type: ignore` instead of casting to Any ?

graphrag/index/flows/create_final_communities.py
```diff
@@ -4,82 +4,72 @@
...
-    filtered["id"] = filtered["community"].apply(lambda _x: str(uuid4()))
-    filtered["community"] = filtered["community"].astype(int)
-    filtered["human_readable_id"] = filtered["community"]
-    filtered["title"] = "Community " + filtered["community"].astype(str)
-
-    # Add period timestamp to the community reports
-    filtered["period"] = datetime.now(timezone.utc).date().isoformat()
+    # join it all up and add some new fields
+    communities = all_grouped.merge(entity_ids, on="community", how="inner")
+    communities["id"] = communities["community"].apply(lambda _x: str(uuid4()))
```
andresmor-ms, line 64: You can remove the apply by using:
```
communities["id"] = [uuid4().hex for _ in range(len(communities)]
```

graphrag/index/flows/create_final_communities.py
```diff
@@ -4,82 +4,72 @@
...
+    communities["human_readable_id"] = communities["community"]
+    communities["title"] = "Community " + communities["community"].astype(str)
 
-    # Add size of the community
-    filtered["size"] = filtered.loc[:, "entity_ids"].apply(lambda x: len(x))
+    # add fields for incremental update tracking
+    communities["period"] = datetime.now(timezone.utc).date().isoformat()
+    communities["size"] = communities.loc[:, "entity_ids"].apply(lambda x: len(x))
```
andresmor-ms, line 70: You can pass functions to the apply, no need for the lambda here, you can do:
```
communities["size"] = communities.loc[:, "entity_ids"].apply(len)
```

graphrag/index/flows/create_final_communities.py
natoverse: Yeah, it complains that a list[str] is not correct for columns. Elsewhere we've done the cast so I stuck with that, but ignore is probably cleaner anyway

graphrag/index/flows/create_final_communities.py
```diff
@@ -4,82 +4,71 @@
...
-        clusters["level_x"] == clusters["level_y"]
-    ].reset_index(drop=True)
-
-    cluster_relationships = (
-        combined_clusters.groupby(["cluster", "level_x"], sort=False)
-        .agg(
-            relationship_ids=("id_y", "unique"),
-            text_unit_ids=("source_id_x", "unique"),
-            entity_ids=("id_x", "unique"),
+    for level in range(max_level + 1):
```
AlonsoGuevara, line 30: If we start from 0 here we can effectively filter out all the -1 communities.
I'm currently filtering them out on node prep for community reports as well.

natoverse: This is starting from level 0. You may be confusing the community ID with the level? IDs can be -1 for nodes when they have no community home at a given level.

AlonsoGuevara (approved):

PR #1510, opened 2024-12-12
Flow cleanup

natoverse:
A few things to reorganize the flows for better and more direct activities. Also renames some properties (such as entity name -> title) to eliminate confusion and later renames.


graphrag/index/flows/create_base_text_units.py
```diff
@@ -74,19 +70,7 @@ async def create_base_text_units(
...
-    if snapshot_transient_enabled:
-        await snapshot(
-            output,
-            name="create_base_text_units",
-            storage=storage,
-            formats=["parquet"],
-        )
-
-    return output
+    return cast("pd.DataFrame", chunked[chunked["text"].notna()].reset_index(drop=True))
```
AlonsoGuevara, line 73: We can avoid this cast btw. I'll make sure to submit a PR following this one

graphrag/index/operations/extract_entities/strategies/graph_intelligence.py
```diff
@@ -106,7 +106,7 @@ async def run_extract_entities(
             )
 
     entities = [
-        ({"name": item[0], **(item[1] or {})})
+        ({"title": item[0], **(item[1] or {})})
```
AlonsoGuevara, line 109: WIll this change require a re-index?

natoverse: Nope! I just moved it all the way to the original source instead of renaming in the final dataframe, so it's less confusing to track

jgbradley1 (approved):

PR #1554, opened 2024-12-24
Simplify flow config

natoverse:
Simplifies config by working backward from the flow to flatten nested objects and remove properties that are never set beyond defaults. Uses strong types at the lowest levels to avoid error-prone dicts.


graphrag/index/operations/chunk_text/chunk_text.py
```diff
@@ -12,20 +12,22 @@
     progress_ticker,
 )
 
+from graphrag.config.models.chunking_config import ChunkingConfig, ChunkStrategyType
```
jgbradley1, line 15: Minor suggestion. Should this be named `ChunkConfig` so the tense matches the other chunk-related models?

graphrag/config/models/cluster_graph_config.py
```diff
@@ -14,15 +14,10 @@ class ClusterGraphConfig(BaseModel):
...
         description="The maximum cluster size to use.", default=defs.MAX_CLUSTER_SIZE
     )
-    strategy: dict | None = Field(
-        description="The cluster strategy to use.", default=None
+    use_lcc: bool = Field(
+        description="Whether to use the largest connected component.",
+        default=defs.USE_LCC,
+    )
+    seed: int | None = Field(
+        description="The seed to use for the clustering.", default=None
```
jgbradley1, line 22: Should the default seed value be set here instead?

graphrag/config/create_graphrag_config.py
```diff
@@ -522,8 +525,12 @@ def hydrate_parallelization_params(
             )
 
         with reader.use(values.get("cluster_graph")):
+            use_lcc = reader.bool("use_lcc")
             cluster_graph_model = ClusterGraphConfig(
-                max_cluster_size=reader.int("max_cluster_size") or defs.MAX_CLUSTER_SIZE
+                max_cluster_size=reader.int("max_cluster_size")
+                or defs.MAX_CLUSTER_SIZE,
```
jgbradley1, line 531: Can the default value here be defined in the config pydantic model directly instead?

graphrag/index/operations/chunk_text/chunk_text.py
natoverse: I like that idea, will review

graphrag/config/create_graphrag_config.py
natoverse: It is set in the pydantic model as well. I'm not yet sure how we will remove the redundancy, but we need to do it everywhere. Hoping Derek has a good solution

graphrag/config/models/cluster_graph_config.py
natoverse: Good idea - I think the most reasonable approach is actually that not passing any seed results in a new random one each time, and our system can set a default for own repeatability reasons.

graphrag/index/operations/chunk_text/chunk_text.py
natoverse: Ok, looking closer it seems most of our config does have the active tense ("embed_graph", "cluster_graph", etc.). I'll leave it for now but we can revist.

graphrag/config/create_graphrag_config.py
```diff
@@ -318,13 +318,16 @@ def hydrate_parallelization_params(
             reader.envvar_prefix(Section.node2vec),
             reader.use(values.get("embed_graph")),
         ):
+            use_lcc = reader.bool("use_lcc")
```
AlonsoGuevara, line 321: Wonder if we should start just defaulting these and removing them from config. (Explicitly defaulting in this section)

AlonsoGuevara (approved):

graphrag/config/create_graphrag_config.py
natoverse: Some of these I have been removing, but this one was kind of weird. It seems perfectly reasonable to allow users to specify whether to only use the lcc, but to my knowledge we always set this to True. Maybe a good question for Ha

ha2trinh: I'm not sure if any users actually change this value, but for lazygraphrag I actually set it to False by default (so that we don't lose nodes).

natoverse: Ok, I will leave the config in place for now given a False use case

PR #1565, opened 2024-12-28
Add new basic pipeline runner

natoverse:
WIP to remove DataShaper for a simpler pipeline runner


graphrag/api/index.py
```diff
@@ -27,6 +28,7 @@ async def build_index(
     memory_profile: bool = False,
     callbacks: list[WorkflowCallbacks] | None = None,
     progress_logger: ProgressLogger | None = None,
+    use_new_pipeline: bool = False,
```
AlonsoGuevara, line 31: This is for backwards compat of for the draft purpose?

natoverse: just for draft so you can run either way

graphrag/index/workflows/default_workflows.py
```diff
@@ -91,3 +129,18 @@
     create_final_communities: build_create_final_communities_steps,
     generate_text_embeddings: build_generate_text_embeddings_steps,
 }
+
+basic_workflows: dict[str, Any] = {
```
dworthen, line 133: Could we type the value of the dict as `Callable[[GraphRagConfig, PipelineRunContext, VerbCallbacks], Awaitable[None]]` instead of `Any` since they all have the same signature?

natoverse:
This PR was a draft for review before removing the DataShaper code. It has been subsumed by the completed #1581

PR #1594, opened 2025-01-08
allow httpx 0.27

ex0ns:
## Description

A lot of the libraries in the python ecosystem are still locked on 0.27, creating all kind of dependency conflicts. As far as I could tell by overriding the version locally, there should not be any reason for GraphRAG to require httpx version 0.28.

I was wondering if it would be possible to make the requirements of httpx less strict for GraphRAG

## Proposed Changes

Allow version 0.27.X and 0.28.X version of httpx.

## Checklist

## Additional Notes

I ran `poetry run poe test_unit` but did not ran the integration or smoke tests because of the requirement for a OPENAI_API_KEY and GRAPHRAG_API_KEY.

### DIFF
--- a/.semversioner/next-release/patch-20250108184901555481.json
+++ b/.semversioner/next-release/patch-20250108184901555481.json
@@ -1,0 +1,4 @@
+{
+  "type": "patch",
+  "description": "support vers 0.27 of httpx"
+}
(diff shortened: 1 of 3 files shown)


ex0ns:

PR #1644, opened 2025-01-21
Multi-index querying for API layer

KennyZhang1:
## Description

This PR adds support for local and global multi-index querying to the graphrag query API.

## Related Issues

N/A

## Proposed Changes

- Implemented two new functions in the graphrag query API layer for multi index global and local search
- Cleaned up graphrag API by moving helper functions to a separate util file
- Added support for passing in lists of vector stores via the config file

## Checklist

## Additional Notes

- This PR is a first step for integrating multi-index querying into the graphrag library. Once the config file changes and config loading methods are finalized, a separate PR will be opened for integrating multi-index querying at the CLI layer.


graphrag/api/query.py
```diff
@@ -198,13 +201,177 @@ async def global_search_streaming(
...
-            context_data = _reformat_context_data(stream_chunk)  # type: ignore
+            context_data = reformat_context_data(stream_chunk)  # type: ignore
             yield context_data
             get_context_data = False
         else:
             yield stream_chunk
 
 
+@validate_call(config={"arbitrary_types_allowed": True})
+async def multi_global_search(
```
jgbradley1, line 212: In order to not confuse people about which API functions should be used, we should make changes to the naming convention used here. I don't think we need multiple global_search methods at the API layer - we just want one.

Can `multi_global_search()` handle an index list that contains just one index? It appears so at first glance.

If this is the case, I recommend renaming the existing `global_search()` function to `_global_search()` to indicate it's meant to be a private function and rename `multi_global_search()` to `global_search()`.

Or perhaps we move the existing `global_search()` function into `utils/api.py` to keep `api/query.py` clean and readable by external library users.

This comment also applies to the other query strategies in this file as well (local, drift).

dorbaker: Kenny and I had talked about that... There are some parameters for multi_global_search that are not required by standard global_search, such as a list of index names. We could move functions in the utils file and have a global_search be a wrapper that would choose which function based on the input type, with lists going to multi_global_search (even if there's only index) and dataframes going to the current global_search function.

natoverse (approved):

graphrag/config/models/vector_store_config.py
```diff
@@ -61,12 +67,16 @@ def _validate_url(self) -> None:
         default=None,
     )
 
-    container_name: str = Field(
-        description="The database name to use.",
+    container_name: str | list[str] = Field(
```
natoverse, line 70: This goes back to a str only now with the root dict?

graphrag/config/models/vector_store_config.py
```diff
@@ -61,12 +67,16 @@ def _validate_url(self) -> None:
...
+        description="The container name to use.",
         default=defs.VECTOR_STORE_CONTAINER_NAME,
     )
 
-    overwrite: bool = Field(
+    database_name: str | None = Field(
+        description="The database name to use when type == cosmos_db.", default=None
+    )
+
+    overwrite: bool | list[str] = Field(
```
natoverse, line 79: This goes back to a bool only now with the root dict?

graphrag/config/models/graph_rag_config.py
```diff
@@ -224,20 +224,20 @@ def _validate_update_index_output_base_dir(self) -> None:
     )
     """The basic search configuration."""
 
-    vector_store: VectorStoreConfig = Field(
-        description="The vector store configuration.", default=VectorStoreConfig()
+    vector_store: dict[str, VectorStoreConfig] = Field(
+        description="The vector store configuration.",
+        default={"default": VectorStoreConfig()},
```
natoverse, line 229: It looks like you used "output" as the default in defaults.py, so I would use that variable here. I'd probably make sure the default name "default" isn't used anywhere since it could conflict with keywords (I'm thinking maybe JSON versions of settings because `default` is a keyword in JavaScript?)

natoverse (changes requested):

natoverse (approved):

dworthen (approved):

PR #1645, opened 2025-01-21
Fix DRIFT search on Azure AI Search

natoverse:
AZAIS embeddings were not including the vector itself in the list of retrievable fields, which DRIFT relies on for local cosine similarity checks.

This also adds DRIFT and Basic search queries to the smoke tests, so we should catch search issues more reliably with AZAIS and lancedb.


AlonsoGuevara (approved):

natoverse:
Edit: removed the new DRIFT smoke test because it is brittle. We will address that in a separate PR to unblock patching DRIFT with this bugfix.

AlonsoGuevara (approved):

Issue #1506, opened 2024-12-12
[Bug]: Incorrect Handling of overlap Parameter in Chunking Process (with overlap: 0)


jose-mendez-santos:
### Do you need to file an issue?

- [x] I have searched the existing issues and this bug is not already filed.
- [x] My model is hosted on OpenAI or Azure. If not, please look at the "model providers" issue and don't file a new one here.
- [x] I believe this is a legitimate bug, not just a question. If this is a question, please use the Discussions area.

### Describe the bug

When processing a document (input) and splitting it into chunks, the library does not respect the configuration specified in the settings.yaml file if the overlap is set to 0. Instead, it defaults to a value of 100. However, when the overlap is set to 1, the expected behavior is observed. This means there is currently no way to split the original text without applying overlap.

### Steps to reproduce

The problematic configuration in the settings.yaml file is as follows:

chunks:
  size: 1200
  overlap: 0
  group_by_columns: [id]
  encoding_model: o200k_base

### Expected Behavior

The input document has a total of 5,273 tokens. As shown in the image below, the total sums to 5,673, which is 400 tokens higher.

![Image](https://github.com/user-attachments/assets/398f7ebf-e32e-45dc-8986-f61183709575)

When using an overlap of 1, the result is 5,277 tokens, only 4 tokens above the original text. This demonstrates that the overlap is functioning correctly for values greater than 0. However, it does not work when set to 0, meaning when no overlap is desired.

![Image](https://github.com/user-attachments/assets/2d7721c1-665e-4ec2-b470-2a28b7adef7f)

### GraphRAG Config Used

```yaml
chunks:
  size: 1200
  overlap: 0
  group_by_columns: [id]
  encoding_model: o200k_base

```

### Additional Information

- GraphRAG Version: 1.0.0
- Operating System: Windows 11
- Python Version: Python 3.11.9
- Related Issues:

dayesouza:
Hi @jose-mendez-santos 
I was testing it here and I believe this was solved by #1593 
Could you try again please?

PR #1646, opened 2025-01-21
fix config for reading chunk_overlap value

dayesouza:
## Description

This PR fix a bug that when reading the param values, if chunk overlap was 0, it would always fallback to the default value.

## Related Issues

#1506 

## Proposed Changes

- Change of method for reading overlap value to verify that only if it's none it would set the default value

- Created necessary unit tests to check the appropriate config is set when overlap value is 0

## Checklist


dayesouza:
Solved by #1593

PR #1652, opened 2025-01-22
NLP graph extraction

natoverse:
Adds an NLP-based option for graph extraction. This is intended to replace the unmaintaned "nltk" strategy within the existing extraction. Adds a new CLI param "method" to specify this technique.

This particular implementation is a work-in-progress hybrid based on some early NLP-based approaches we called "FastGraphRAG". It has elements of LazyGraphRAG on the indexing side to save costs.


graphrag/api/index.py
```diff
@@ -81,18 +86,40 @@ async def build_index(
...
-        "create_final_entities",
-        "create_final_relationships",
-        "create_final_nodes",
-        "create_final_communities",
-        *(["create_final_covariates"] if config.claim_extraction.enabled else []),
-        "create_final_text_units",
-        "create_final_community_reports",
-        "generate_text_embeddings",
-    ]
+def _get_workflows_list(config: GraphRagConfig, method: IndexingMethod) -> list[str]:
```
AlonsoGuevara, line 89: nit: We could move this to a Factory method and encapsulate that from the index code

natoverse: Kenny moved this util on the multi-vector PR, so I'll address this when I merge main after that one is done.

graphrag/config/defaults.py
```diff
@@ -52,36 +56,67 @@
...
 COMMUNITY_REPORT_MODEL_ID = DEFAULT_CHAT_MODEL_ID
+
+# Graph extraction via LLM
 ENTITY_EXTRACTION_ENTITY_TYPES = ["organization", "person", "geo", "event"]
 ENTITY_EXTRACTION_MAX_GLEANINGS = 1
 ENTITY_EXTRACTION_MODEL_ID = DEFAULT_CHAT_MODEL_ID
+
+# Graph extraction via NLP
+NLP_NORMALIZE_EDGE_WEIGHTS = True
+NLP_EXTRACTOR_TYPE = NounPhraseExtractorType.RegexEnglish
```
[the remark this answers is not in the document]
natoverse: The other two options work on non-English, but are slower so this is the default. I'll make sure to update the docs PR to help with clarity on this.

graphrag/api/index.py
natoverse: I just moved the workflow lists into a little pipeline generator

AlonsoGuevara (dismissed):

graphrag/index/flows/create_final_community_reports_text.py
```diff
@@ -0,0 +1,112 @@
...
+            local_context_df=local_contexts,
+            community_hierarchy_df=community_hierarchy,
+            level=level,
+            max_tokens=summarization_strategy.get(
+                "max_input_tokens", defaults.COMMUNITY_REPORT_MAX_INPUT_LENGTH
+            ),
+        )
+        level_contexts.append(level_context)
+
+    community_reports = await summarize_communities(
```
ha2trinh, line 73: in the FastGraphRAG's prompt we extract date_range as a separate attribute, which we didn't have in the standard GraphRAG. Should we modify this method to parse that value and store in the community report's attributes field?

graphrag/index/operations/summarize_communities_text/prompts.py
```diff
@@ -0,0 +1,93 @@
+# Copyright (c) 2024 Microsoft Corporation.
+# Licensed under the MIT License
+
+"""A file containing prompts definition."""
+
+COMMUNITY_REPORT_PROMPT = """
```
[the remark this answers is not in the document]
natoverse: Yes, but there is a bigger prompt-loading refactor I want to do. Right now the config makes it difficult to swap this particular prompt because I am reusing the summarization code so I had to do a little manual overriding.

graphrag/index/operations/prune_graph.py
```diff
@@ -0,0 +1,92 @@
...
+
+import graphrag.index.operations.build_noun_graph.schemas as schema
+
+if TYPE_CHECKING:
+    from networkx.classes.reportviews import DegreeView
+
+
+def prune_graph(
+    graph: nx.Graph,
+    min_node_freq: int = 1,
```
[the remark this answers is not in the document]
natoverse: I don't think so - I consider these defaults the basic functional defaults with valid values, and the config is our expression of the best params to invoke it with. BTW, our new config ensures all values will be populated properly, so most of these defaults are a courtesy if someone wants to use the function elsewhere. Before we used pydantic validation, these needed to reflect the default config.

graphrag/index/flows/create_final_community_reports_text.py
natoverse: Will add as a separate feature, since this is not present in the reports at all right now

AlonsoGuevara (approved):

ha2trinh (approved):

PR #1681, opened 2025-02-06
add option to add metadata into text chunks

dayesouza:
## Description

change metadata column in input table
add new default column into final_documents called creation_date
add option to prepend metadata into text chunks (default False)
add option to count metadata tokens in max tokens (default False)

## Proposed Changes
* Added new configuration options `prepend_metadata` and `count_tokens_with_metadata` to the `ChunkingConfig` class in `graphrag/config/models/chunking_config.py`. These options allow metadata to be prepended to chunks and included in token counts.

* Updated the `create_base_text_units` function to include the new `prepend_metadata` and `count_tokens_with_metadata` parameters, and adjusted the aggregation logic to handle metadata.

* Modified the `chunk_text` function to support the new metadata options, including updating the `run_strategy` function to prepend metadata and count tokens if specified

* Introduced a new method `get_creation_date` in `BlobPipelineStorage` and `FilePipelineStorage` to retrieve the creation date of files. This method is now used in various parts of the code to add the creation date to data frames. 

* Updated the `load_file` function in `csv.py` and `text.py` to include the file creation date in the data frame.

* Added new integration tests and unit

## Checklist


graphrag/config/models/chunking_config.py
```diff
@@ -26,3 +26,11 @@ class ChunkingConfig(BaseModel):
     encoding_model: str = Field(
         description="The encoding model to use.", default=defs.ENCODING_MODEL
     )
+    prepend_metadata: bool = Field(
+        description="Prepend metadata into each chunk.",
+        default=defs.GRAPHRAG_CHUNK_PREPEND_METADATA,
+    )
+    count_tokens_with_metadata: bool = Field(
```
natoverse, line 33: This naming could be a little clearer. I don't particularly like "tokens" also since the chunking may not be tokens. What about "chunk_size_includes_metadata"?

graphrag/index/flows/create_base_text_units.py
```diff
@@ -32,15 +34,20 @@ def create_base_text_units(
...
             if len(group_by_columns) > 0
             else sort.groupby(lambda _x: True)
         )
-        .agg(texts=("text_with_ids", list))
+        .agg(agg_dict)
         .reset_index()
     )
+    aggregated.rename(columns={"text_with_ids": "texts"}, inplace=True)
 
     aggregated["chunks"] = chunk_text(
```
natoverse, line 52: Do this as two steps so we don't need to edit the chunk_text method, something like:

`aggregated["chunks"] = chunk_text(...)`

```
if prepend_metadata:
  aggregated["chunks"] = aggregated["chunks"].apply(lambda x: meta + x)
```

natoverse (changes requested):

natoverse (changes requested):

graphrag/index/workflows/create_final_documents.py
```diff
@@ -17,16 +17,15 @@
 
 
 async def run_workflow(
-    config: GraphRagConfig,
+    _: GraphRagConfig,
```
natoverse, line 20: name this `_config`

natoverse (approved):

tests/unit/indexing/operations/chunk_text/test_strategies.py
```diff
@@ -84,3 +88,40 @@ def test_non_string_input(self, mock_get_encoding):
         # Verify non-string input is handled
         assert len(chunks) > 0
         assert "123" in chunks[0].text_chunk
+
+
+@patch("tiktoken.get_encoding")
+def test_get_encoding_fn_encode(mock_get_encoding):
```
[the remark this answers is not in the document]
dayesouza: In this case, since we only want to test that the function is working and it's being called, I mock instead of using the library because we don't really care about the return

natoverse (approved):

PR #1721, opened 2025-02-19
Query callbacks

natoverse:
Adds callbacks to the query methods, allowing an alternative way of tracking results (in particular, using on_llm_new_token for frameworks that don't work well with AsyncGenerators)


graphrag/query/structured_search/local_search/search.py
```diff
@@ -150,11 +154,12 @@ async def stream_search(
             {"role": "user", "content": query},
         ]
 
-        # send context records first before sending the reduce response
-        yield context_result.context_records
+        if self.callbacks:
```
AlonsoGuevara, line 157: Nit: I think the if may not be necessary, as the for in won't yield any result when None

AlonsoGuevara (approved):

AlonsoGuevara (approved):

Issue #1825, opened 2025-03-19
[Bug]: Prompt Tunning client does not comply with config embedding batch size


nievespg1:
### Do you need to file an issue?

- [x] I have searched the existing issues and this bug is not already filed.
- [x] My model is hosted on OpenAI or Azure. If not, please look at the "model providers" issue and don't file a new one here.
- [x] I believe this is a legitimate bug, not just a question. If this is a question, please use the Discussions area.

### Describe the bug

When running prompt tune using the automatic selection method, the system will attempt to embed all the text chunks within one request no matter the size of the payload. 

By default, the batch-size should not be larger than 16 text chunks and the token count should be below 8191 for the whole batch. 

### Steps to reproduce

```bash
root_dir="/your/root/dir"
graphrag prompt-tune --verbose --logger rich \
    --root $root_dir \
    --output $root_dir/prompts \
    --config $root_dir/settings/settings.yaml \
    --selection-method auto \
    --n-subset-max 1000 \
    --k 100 \
    --chunk-size 800 \
    --overlap 0 \
    --max-tokens 2000 \
    --min-examples-required 2 \
    --no-discover-entity-types
```

### Expected Behavior

Given the large value of `n-subset-max` and `k` the system will try and push a very large list of text which easily triggers 429 errors.  

### GraphRAG Config Used

```yaml
# Paste your config here

```

### Logs and screenshots

Below is an image of my embedding's endpoint. I ran 4 prompt tunning tuning jobs and every time the system would try to embed 698,463 tokens in 1 request.

![Image](https://github.com/user-attachments/assets/7fa2114b-0cfa-4e4c-9d93-87f573471d13)

### Additional Information

- GraphRAG Version: latest/main
- Operating System: 22.04.1-Ubuntu SMP 
- Python Version: Python 3.11.11
- Related Issues: N/A

PR #1826, opened 2025-03-19
Gnievesponce prompt tune embedd chunking

nievespg1:
## Description

When running prompt tune using the automatic selection method, the system will attempt to embed all the text chunks within one request no matter the size of the payload.

By default, the batch-size should not be larger than 16 text chunks and the token count should be below 8191 for the whole batch.

## Related Issues

#1825

## Proposed Changes

Modify [graphrag/prompt_tune/loader/input.py](https://github.com/microsoft/graphrag/blob/main/graphrag/prompt_tune/loader/input.py) to add logic that chunks large embeddings jobs/request similarly to how we do it in the indexing workflow. Here is an example workflow with a correct batching strategy: [graphrag/index/operations/embed_text/strategies/openai.py](https://github.com/microsoft/graphrag/blob/main/graphrag/index/operations/embed_text/strategies/openai.py)

## Checklist

## Additional Notes

No additional notes


graphrag/api/prompt_tune.py
```diff
@@ -111,7 +111,7 @@ async def generate_indexing_prompts(
 
     # if max_retries is not set, inject a dynamically assigned value based on the number of expected LLM calls
     # to be made or fallback to a default value in the worst case
-    if default_llm_settings.max_retries == -1:
+    if default_llm_settings.max_retries < -1:
```
AlonsoGuevara, line 114: -1 means the max retries will be dynamically calculated.
When would it be less than -1?

nievespg1: It should never be less than one. This was just some logic to address an edge case where the state is "undefined"; like a typo for example. 

Should we raise an exception when the value of `max_retries` is out of bounds? Or should we assume that any value lower than 0 means calculate `max_retries` dynamically?

AlonsoGuevara: I would like that we iterate over this. This should be handed and resolved when the config is loaded.

Approving since this change is required for experimentation

AlonsoGuevara (approved):

graphrag/api/prompt_tune.py
nievespg1: Agree.

PR #1855, opened 2025-03-31
fix: Use format() for community report prompt

zjx20:
## Description

This pull request addresses two key issues in the `CommunityReportsExtractor` related to prompt formatting:

1.  **Incorrect Handling of Escaped Braces:** The previous implementation used `replace()` to inject the input text into the prompt. However, the prompt template contains escaped curly braces (`{{` and `}}`) that are intended to be treated as literal characters, not as placeholders for substitution. `replace()` does not handle these escaped braces correctly, leading to malformed prompts.
https://github.com/microsoft/graphrag/blob/b7b2b562cea7893369442a7de07032360d7cd6c3/graphrag/prompts/index/community_report.py#L22-L37
3.  **Inconsistency:** Other parts of the codebase use the `.format()` method for prompt construction. The `CommunityReportsExtractor` was an outlier, using `replace()` instead. This inconsistency made the code harder to understand and maintain.

This PR resolves these issues by switching from `replace()` to `.format()` for prompt construction.

## Related Issues

## Proposed Changes

-   Modified the `__call__` method in `CommunityReportsExtractor` to use `.format()` instead of `.replace()` when constructing the prompt.
-   The prompt now correctly handles escaped curly braces (`{{` and `}}`) within the template.
-   The prompt now uses the format method to inject the input text into the prompt.

## Checklist

## Additional Notes

### DIFF
--- a/graphrag/index/operations/summarize_communities/community_reports_extractor.py
+++ b/graphrag/index/operations/summarize_communities/community_reports_extractor.py
@@ -74,9 +74,9 @@ class CommunityReportsExtractor:
         output = None
         try:
             input_text = inputs[self._input_text_key]
-            prompt = self._extraction_prompt.replace(
-                "{" + self._input_text_key + "}", input_text
-            )
+            prompt = self._extraction_prompt.format(**{
+                self._input_text_key: input_text
+            })
             response = await self._model.achat(
                 prompt,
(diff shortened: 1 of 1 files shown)


natoverse:
Nice catch. Ironically, I made the same fix while working on the reasoning model support! We're just checking that all the results are ok with the minor prompt modifications that we made, and then it will hit main.

PR #1888, opened 2025-04-22
NLP graph parity

natoverse:
Improves our NLP extraction and aligns with new approaches we've been testing


graphrag/index/operations/build_noun_graph/build_noun_graph.py
```diff
@@ -98,7 +99,7 @@ def _extract_edges(
         text_units_df.groupby("text_unit_id").agg({"title": list}).reset_index()
     )
     text_units_df["edges"] = text_units_df["title"].apply(
-        lambda x: _create_relationships(x)
+        lambda x: list(combinations(x, 2))
```
AlonsoGuevara, line 102: I think we could do something like:

```python
titles = text_units_df["title"].tolist()
all_edges = [ list(combinations(t, 2)) for t in titles]

text_units_df = text_units_df.assign(edges=all_edges)
```
to get rid of the apply. May need to testing tho, but I think it could work.

AlonsoGuevara: Andres just ran this on a branch to check speed and seems a bit slower than the apply (for the dataset he used).
We are checking for other ways to make it faster and non-apply.

andresmor-ms: I tested this on my local branch and it works

graphrag/index/operations/build_noun_graph/build_noun_graph.py
```diff
@@ -122,7 +123,7 @@ def _extract_edges(
         strict=False,
     )
```
andresmor-ms, line 124: This code can be replaced by this, to remove the applies used (also tested locally on my branch)

```python
edge_df[["source", "target"]] = edge_df["edges"].to_list()
edge_df["min_source"] = edge_df[["source", "target"]].min(axis=1)
edge_df["max_target"] = edge_df[["source", "target"]].max(axis=1)
edge_df = edge_df.drop(columns=["source", "target"]).rename(columns={"min_source": "source", "max_target": "target"})

edge_df = edge_df[(edge_df.source.notna()) & (edge_df.target.notna())]
edge_df = edge_df.drop(columns=["edges"])
```

graphrag/index/operations/build_noun_graph/build_noun_graph.py
```diff
@@ -94,64 +95,42 @@ def _extract_edges(
...
-        text_units_df.groupby("text_unit_id").agg({"title": list}).reset_index()
+        text_units_df.groupby("text_unit_id")
+        .agg({"title": lambda x: list(x) if len(x) > 1 else np.nan})
+        .reset_index()
     )
-    text_units_df["edges"] = text_units_df["title"].apply(
-        lambda x: _create_relationships(x)
+    text_units_df = text_units_df.dropna()
+    titles = text_units_df["title"].tolist()
+    all_edges: Any = [list(combinations(t, 2)) for t in titles]
```
andresmor-ms, line 106: if you want to remove that Any I believe you can use:
list[list[tuple[str, str]]]

AlonsoGuevara (approved):

AlonsoGuevara (approved):

PR #1890, opened 2025-04-22
Optional embeddings

natoverse:
Makes all tables optional in the embeddings workflow, so that if tables are missing (e.g., due to intentional partial workflows) they only fail if an embedded field from that table is expected.


graphrag/index/workflows/generate_text_embeddings.py
```diff
@@ -35,13 +39,23 @@ async def run_workflow(
...
-    community_reports = await load_table_from_storage(
-        "community_reports", context.storage
-    )
+    documents = None
+    relationships = None
+    text_units = None
+    entities = None
+    community_reports = None
+    if await storage_has_table("documents", context.storage):
+        documents = await load_table_from_storage("documents", context.storage)
```
AlonsoGuevara, line 48: What if load_table_from_storage returns a null if not exists? To avoid all the if assignments

natoverse: this catches the null with the has_table check

AlonsoGuevara (approved):

PR #1892, opened 2025-04-23
Chunking fix

natoverse:
Fixes a minor off-by-one-ish logical error in text chunking that allowed for non-useful final chunks.


ha2trinh (approved):

natoverse:
Abandoning - didn't realize these fixes were already on Ha's query fix branch, which is a separate PR

Issue #1900, opened 2025-04-26
[Issue]: KeyError: "title" when generating community reports using build_index()


droideronline:
### Do you need to file an issue?

- [x] I have searched the existing issues and this bug is not already filed.
- [x] My model is hosted on OpenAI or Azure. If not, please look at the "model providers" issue and don't file a new one here.
- [x] I believe this is a legitimate bug, not just a question. If this is a question, please use the Discussions area.

### Describe the issue

When running build_index using the Python API (Graphrag v2.2.0, Python 3.11.0), the process fails during community report extraction due to a missing "title" key in the prompt template formatting

### Steps to reproduce

1. Set up Graphrag v2.2.0 with Python 3.11.
2. Run build_index() function via Python API.
3. Observe error when creating community reports.

### GraphRAG Config Used

```yaml
### This config file contains required core defaults that must be set, along with a handful of common optional settings.
### For a full list of available settings, see https://microsoft.github.io/graphrag/config/yaml/

### LLM settings ###
## There are a number of settings to tune the threading and token limits for LLM calls - check the docs.

models:
  default_chat_model:
    type: azure_openai_chat  # or openai_chat
    api_base: https://xxxxx.com
    api_version: 2023-07-01-preview
    auth_type: api_key # or azure_managed_identity
    api_key: ${GRAPHRAG_API_KEY} # set this in the generated .env file
    # audience: "https://cognitiveservices.azure.com/.default"
    # organization: <organization_id>
    model: gpt-4-turbo-preview
    deployment_name: gpt-4o-mini-2024-07-18 # deployment name is modal name
    # encoding_model: cl100k_base # automatically set by tiktoken if left undefined
    model_supports_json: true # recommended if this is available for your model.
    concurrent_requests: 25 # max number of simultaneous LLM requests allowed
    async_mode: threaded # or asyncio
    retry_strategy: native
    max_retries: -1                   # set to -1 for dynamic retry logic (most optimal setting based on server response)
    tokens_per_minute: 0              # set to 0 to disable rate limiting
    requests_per_minute: 0            # set to 0 to disable rate limiting
  default_embedding_model:
    type: azure_openai_embedding # or openai_embedding
    api_base: https://xxxxx.com
    api_version: 2023-07-01-preview
    auth_type: api_key # or azure_managed_identity
    api_key: ${GRAPHRAG_API_KEY}
    # audience: "https://cognitiveservices.azure.com/.default"
    # organization: <organization_id>
[...]

PR #1901, opened 2025-04-26
Fix: Replace manual placeholders to avoid .format() issues in COMMUNITY_REPORT_PROMPT

droideronline:
## Description

- In the `community_reports_extractor` file Using .format() caused errors because the COMMUNITY_REPORT_PROMPT has other `{}` braces (e.g., in JSON examples).
- Python .format() tries to parse those and fails.

## Related Issues

#1900 

## Proposed Changes

- Switched to using .replace() to manually substitute `{input_text}` and `{max_report_length}`.
- Avoids conflict with embedded braces and ensures prompt generation works correctly.

## Checklist

## Additional Notes

- The error occurred because .format() was used on a prompt where only the input variables ({input_text} and {max_report_length}) were meant to be replaced.
- Although all non-input curly braces (for JSON structure) were correctly escaped with double brackets {{ }}, using .format() still caused issues due to complex formatting.
- Replaced .format() with manual .replace() for input_text and max_report_length to fix the problem safely without impacting the rest of the prompt structure.

### DIFF
--- a/graphrag/index/operations/summarize_communities/community_reports_extractor.py
+++ b/graphrag/index/operations/summarize_communities/community_reports_extractor.py
@@ -73,10 +73,11 @@ class CommunityReportsExtractor:
         """Call method definition."""
         output = None
         try:
-            prompt = self._extraction_prompt.format(**{
-                INPUT_TEXT_KEY: input_text,
-                MAX_LENGTH_KEY: str(self._max_report_length),
-            })
+            prompt = self._extraction_prompt.replace(
+                "{" + INPUT_TEXT_KEY + "}", input_text
+            ).replace(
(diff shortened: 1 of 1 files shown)


droideronline:
@AlonsoGuevara  Please review the PR when you have time. Thanks!

natoverse:
Did you run prompt tuning? It looks like the issue is that prompt tune calls format to insert its variables, reducing the bracket count by one when written back out. Running with default prompts does not exhibit the issue.

We're going to look for a fix on the prompt tuning side so we can retain the use of `format`

droideronline:
Yes, I did manual prompt tuning, understood now it makes sense. Thanks

PR #1908, opened 2025-04-30
Update Index as workflows

andresmor-ms:
## Description

Use workflows to update the results from updating the index.

## Proposed Changes

- Remove update_dataframe_outputs in favor of splitting it into several workflows
- Use is_update_run to add update workflows in the workflow list
- Add timestamp to the context to manage the different storages used during the update process

## Checklist

- [X] I have updated the documentation (if necessary).
- [X] I have added appropriate unit tests (if applicable).


graphrag/index/workflows/factory.py
```diff
@@ -66,4 +83,5 @@ def _get_workflows_list(
                 "create_final_text_units",
                 "create_community_reports_text",
                 "generate_text_embeddings",
+                *(update_workflows if is_update_run else []),
```
natoverse, line 86: FGR does not do covariates, so you may want to test if this is ok

natoverse: I see below you have a guard, so this could probably be ignored

andresmor-ms: the covariates workflow will just do nothing if there's no covariates :)

graphrag/index/run/run_pipeline.py
```diff
@@ -62,48 +61,47 @@ async def run_pipeline(
...
+            storage=storage, cache=cache, callbacks=callbacks, state=state
+        )
+
         async for table in _run_pipeline(
             pipeline=pipeline,
             config=config,
             dataset=dataset,
-            cache=cache,
-            storage=storage,
             callbacks=callbacks,
```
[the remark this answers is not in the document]
andresmor-ms: Removed :)

PR #1932, opened 2025-05-15
Various minor updates

natoverse:
Hodgepodge of things to streamline use in various projects

- updated the embeddings config/flow to be simpler and more resilient to experimental config
- added missing text unit IDs to the Community model and loader
- added a handful of graph utilities


graphrag/index/operations/cluster_graph.py
```diff
@@ -6,6 +6,7 @@
 import logging
 
 import networkx as nx
+from graspologic.partition import hierarchical_leiden
```
AlonsoGuevara, line 9: Did you run from cli just to check time impact of this change?

AlonsoGuevara (approved):

PR #1957, opened 2025-05-30
Dorbaker/miq fix

dorbaker:
## Description

Multi index query was broken. These updates fix the issues. In addition to fixing the code, the multi index query example notebook has also been updated.

## Checklist


natoverse:
We put in a patch with #2063 since we let this diverge from main

PR #2009, opened 2025-07-23
Added index support for in memory csv files

snehitgajjar:
## Description

GraphRAG current support input for indexing using local file directory and azure resources. We have a scenario where we cannot use any of those sources as input. So, I have added changes that allows user to provide list of csv(DataFrame) files as an input to build_index function and uses those files instead of trying to read from input directories or databases.

## Related Issues
N/A

## Proposed Changes

* Added a new file loader called memory_csv that uses the user provided list of files and processes it.
* Updated necessary class/functions to pass the list of files to new file loader.

## Checklist

- [ ] I have updated the documentation (if necessary). I will update it after initial review from graphrag team
- [ ] I have added appropriate unit tests (if applicable). I will add it after initial review from graphrag team

## Additional Notes

### DIFF
--- a/docs/examples_notebooks/api_overview.ipynb
+++ b/docs/examples_notebooks/api_overview.ipynb
@@ -21,7 +21,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 5,
    "metadata": {},
    "outputs": [],
    "source": [
@@ -37,11 +37,11 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 5,
    "metadata": {},
    "outputs": [],
    "source": [
-    "PROJECT_DIRECTORY = \"<your project directory>\""
+    "PROJECT_DIRECTORY = \"C:\\\\Users\\\\snehitgajjar\\\\ragtest\""
    ]
   },
   {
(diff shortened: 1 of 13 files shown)


docs/examples_notebooks/api_overview.ipynb
```diff
@@ -21,7 +21,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 5,
```
andresmor-ms, line 24: Please clean the output of the notebook

graphrag/index/input/util.py
```diff
@@ -53,6 +53,39 @@ async def load_files(
     return result
 
 
+def load_files_from_memory(
+    loader: Any,
+    config: InputConfig,
+    input_files: list[pd.DataFrame] | None = None,
+) -> pd.DataFrame:
+    """Load files from storage and apply a loader function."""
+    if not input_files or len(input_files) == 0:
```
andresmor-ms, line 62: you should be able to remove the `or len(input_files) == 0` since in python empty lists are falsy

docs/examples_notebooks/api_overview.ipynb
snehitgajjar: Oh, this was added by mistake. I will clean it up before I publish it.

graphrag/index/input/memory_csv.py
```diff
@@ -0,0 +1,51 @@
...
+) -> pd.DataFrame:
+    """Load csv inputs from a directory."""
+    logger.info("Loading csv files from %s", config.storage.base_dir)
+
+    if input_files is not None and len(input_files) == 0:
+        error_msg = "input_files is an empty list, no files to load."
+        raise ValueError(error_msg)
+
+    def load_file(
+        group: dict | None, input_file: pd.DataFrame = pd.DataFrame()
```
andresmor-ms, line 30: wouldn't it be better to have input_file first and required and then group as optional so that in load_files_from_memory you can do:

```python
for file in input_files:
    files_loaded.append(loader(file))
```

andresmor-ms: If both are optional i'd make it input_file: pd.DataFrame | None = None and do:

```py
data = input_file or pd.DataFrame()
```

PR #2010, opened 2025-07-24
Use in memory dictionary as an index input

snehitgajjar:
## Description

GraphRAG current support input for indexing using local file directory and azure resources. We have a scenario where we cannot use any of those sources as input. So, I have added changes that allows user to provide list of csv(DataFrame) files as an input to build_index function and uses those files instead of trying to read from input directories or databases.

## Related Issues

## Proposed Changes

* Updated MemoryPipellineStorage to read input file metadata from dictionary instead of FilePipelineStorage.
* Added support to pass dictionary with filename and DataFrameObject from build_index function.

## Checklist

- [ ] I have updated the documentation (if necessary). I will update it after initial review from graphrag team

## Additional Notes


graphrag/index/input/csv.py
```diff
@@ -25,8 +26,14 @@ async def load_csv(
     async def load_file(path: str, group: dict | None) -> pd.DataFrame:
         if group is None:
             group = {}
-        buffer = BytesIO(await storage.get(path, as_bytes=True))
-        data = pd.read_csv(buffer, encoding=config.encoding)
+
+        if isinstance(storage, MemoryPipelineStorage):
+            # If using memory storage, we can directly read from the in-memory DataFrame
+            data = await storage.get(path)
```
AlonsoGuevara: Adding a not here... Wonder if we could just make storage.get return the byte representation of the dataframe. I know it is redundant but won't require this class check and will keep the same signature.
It is more a nit, feel free to dismiss, it is just that I dislike instance checks.

graphrag/storage/memory_pipeline_storage.py
```diff
@@ -34,7 +52,7 @@ async def get(
         -------
             - output - The value for the given key.
         """
-        return self._storage.get(key)
+        return self._storage["files"].get(key)
```
AlonsoGuevara: We should add a validation that when a storage type is memory, we always have the files property in config

graphrag/storage/memory_pipeline_storage.py
```diff
@@ -63,16 +84,44 @@ async def delete(self, key: str) -> None:
...
+        del self._storage["files"][key]
 
     async def clear(self) -> None:
         """Clear the storage."""
         self._storage.clear()
 
     def child(self, name: str | None) -> "PipelineStorage":
         """Create a child storage instance."""
-        return MemoryPipelineStorage()
+        return self
```
AlonsoGuevara, line 111: What would happen if we used an existing key in the child, under this approach it would override the parent's key

AlonsoGuevara:
Overall comment, changes are breaking the unit tests for storage checks

graphrag/index/input/csv.py
snehitgajjar: I can test that out. I agree it can cleanup this file.

graphrag/storage/memory_pipeline_storage.py
snehitgajjar: The idea is to not have child properties for memory storage. Child property makes sense for file type storage where there is hierarchy. I feel there is no need to have hierarchy here.

graphrag/index/input/csv.py
snehitgajjar: Updated.

PR #2030, opened 2025-08-14
Logging improvements

natoverse:
Resolves a handful of nits introduced by the copilot logging refactor.

### DIFF
--- a/graphrag/callbacks/console_workflow_callbacks.py
+++ b/graphrag/callbacks/console_workflow_callbacks.py
@@ -1,0 +1,45 @@
+# Copyright (c) 2024 Microsoft Corporation.
+# Licensed under the MIT License
+
+"""A logger that emits updates from the indexing engine to the console."""
+
+from graphrag.callbacks.noop_workflow_callbacks import NoopWorkflowCallbacks
+from graphrag.index.typing.pipeline_run_result import PipelineRunResult
+from graphrag.logger.progress import Progress
+
+
+class ConsoleWorkflowCallbacks(NoopWorkflowCallbacks):
+    """A logger that writes to a console."""
+
+    _verbose = False
+
+    def __init__(self, verbose = False):
+        self._verbose = verbose
+
+    def pipeline_start(self, names: list[str]) -> None:
+        """Execute this callback to signal when the entire pipeline starts."""
+        print("Starting pipeline with workflows:", ", ".join(names))
+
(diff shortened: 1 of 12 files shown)


graphrag/cli/query.py
```diff
@@ -19,8 +18,7 @@
 if TYPE_CHECKING:
     import pandas as pd
 
-# Initialize standard logger
-logger = logging.getLogger(__name__)
+# ruff: noqa: T201
```
AlonsoGuevara, line 21: Is this rule for print() ? Or what check are we ignoring?

AlonsoGuevara: Ah ignore, just saw it below

graphrag/callbacks/console_workflow_callbacks.py
```diff
@@ -0,0 +1,46 @@
...
+"""A logger that emits updates from the indexing engine to the console."""
+
+from graphrag.callbacks.noop_workflow_callbacks import NoopWorkflowCallbacks
+from graphrag.index.typing.pipeline_run_result import PipelineRunResult
+from graphrag.logger.progress import Progress
+
+# ruff: noqa: T201
+
+
+class ConsoleWorkflowCallbacks(NoopWorkflowCallbacks):
```
jgbradley1, line 13: What is the value of `ConsoleWorkflowCallbacks`? If the intent is to use it to print messages to stdout, we can do that with the `logging` module. I intentionally tried to remove this design pattern because it encouraged a bad practice of logging in a way that external systems could not capture messages as easily for telemetry purposes.

The logging module is already setup to print to stdout. If this callback class is needed, can we add the following code snippet to this file
```
import logging
```

and then have the callback class make calls to `logger.info()` instead of calling `print()` directly?

This should have the same desired effect of printing to the console (the current behavior) and allow external systems to capture the same logged message.

natoverse: The issue was that the new logging setup created two logging handlers, one configured by the user with the ReportingConfig, and the other an always-on console logger. This made the CLI extremely verbose with no option to adjust it, when we previously had some simple progress reporting (at the expense of a lot of logger complexity though, which the logging refactor fixed). So I just wanted a very few specific messages logged to the console from the CLI only, hence the console callbacks. The entirety of the system at the API level on down still uses the standard loggers (including the progress ticks).

PR #2034, opened 2025-08-26
Input docs API parameter

natoverse:
This adds an optional input docs param to the indexing API. Several users have requested the ability to skip input finding/parsing and just supply the dataframe directly, so this provides a mechanism to do so.


graphrag/api/index.py
```diff
@@ -91,11 +95,6 @@ async def build_index(
     return outputs
 
 
-def register_workflow_function(name: str, workflow: WorkflowFunction):
```
AlonsoGuevara, line 94: Don't we need this, if we are using GraphRAG as a lib, to register workflows?

natoverse: It's not actually used anywhere, and our examples always just use the PipelineFactory directly instead of the util

PR #2040, opened 2025-09-03
Added Postgres pipeline storage

AICrafterZheng:
## Description
We need a RAG solution for on-prem deployments. The current pipeline storage options (Cosmos DB and Blob Storage) do not support on-prem environments, and we want to avoid storing the graph in files or memory. To address this, I added support for PostgreSQL as a storage option, providing an additional way to persist graphs.

## Related Issues

https://github.com/microsoft/graphrag/discussions/1679

## Proposed Changes

1. Implemented a PostgreSQL storage pipeline, following the same design pattern as `cosmosdb_pipeline_storage.py` and `file_pipeline_storage.py`.
2. Because id serves as the primary key in PostgreSQL tables, it is explicitly included in the graph during initialization.

## Checklist

## Additional Notes


natoverse:
https://github.com/microsoft/graphrag/discussions/2211

PR #2049, opened 2025-09-09
Re-implement hierarchical Leiden

natoverse:
Copies and re-implements pieces from graspologic for Leiden, modularity, and largest connected components to remove python dependency and use graspologic-native directly.


pyproject.toml
```diff
@@ -66,6 +65,8 @@ dependencies = [
     "tqdm>=4.67.1",
     "textblob>=0.18.0.post0",
     "spacy>=3.8.4",
+    "graspologic-native>=1.2.5",
+    "graspologic>=3.4.4",
```
AlonsoGuevara: Do we still need to import graspologic? Can't remember if we had to

natoverse: No, but since I did a separate PR for the node2vec removal, it needs to stay here. However, I have merged that one so I'll update this PR now to remove it completely

PR #2051, opened 2025-09-09
Tokenizer

dworthen:
Remove manual use of tiktoken and replace with Tokenizer interface. Adds support for encoding and decoding the models supported by litellm.

This PR starts to fade out the use/reliance on `LanguageModelConfig.encoding_model` and instead uses a factory method to get a Tokenizer (encoder/decoder) based on the LanguageModel selected model provider:

- A backwards compatible, tiktoken based Tokenizer when using fnllm and
- A LiteLLM based Tokenizer to support all the models supported by litellm, including openai models which in turn just uses tiktoken internally in litellm. 

Users can still specify a tiktoken based  `LanguageModelConfig.encoding_model` for manual control of encoding/decoding, even when using a litellm model provider. For example, user can use an Anthropic model but specify a  `LanguageModelConfig.encoding_model` of `cl100k_base` in which case tiktoken with `cl100k_base` will be used as the Tokenizer throughout even though LLM calls will be made to Anthropic. 

There are still some manual use of tiktoken in the code base:

1. Some fnllm specific utilities that will be removed once fnllm is removed. 
2. Text chunking: The token-based text chunking flows are independent from language model configuration. The `ChunkingConfig` has its own `encoding_model` field. This can be cleaned up to become more unified with the rest of the config by having `ChunkingConfig` specify a `chat_model_id` referencing a registered `LanguageModelConfig` as done elsewhere in the code. 

The above proposed updates are breaking changes and thus saved for `v3` branch. Once the above changes are made in v3 then `LanguageModelConfig.encoding_model` field could be removed entirely if we deem manual control is not necessary.


graphrag/query/context_builder/local_context.py
```diff
@@ -24,12 +23,13 @@
     get_out_network_relationships,
     to_relationship_dataframe,
 )
-from graphrag.query.llm.text_utils import num_tokens
```
AlonsoGuevara, line 27: You could still use num_tokens so we are not doing that many replacements all over the code. And modify in a single place

graphrag/tokenizer/get_tokenizer.py
```diff
@@ -0,0 +1,45 @@
...
+            LiteLLM supports token encoding/decoding for the range of models it supports.
+        encoding_model: str, optional
+            A tiktoken encoding model to use if no model configuration is provided. Only used if a
+            model configuration is not provided.
+
+    Returns
+    -------
+        An instance of a Tokenizer.
+    """
+    if model_config is not None:
```
AlonsoGuevara, line 38: This is closing the code to just tiktoken or litellm, which is bounding us to these implementations. I would suggest using a Factory and protocols. As users can inject their own llm implementations, they could also inject tokenizer implementations as they see fit.

.semversioner/next-release/minor-20250813214258257076.json
```diff
@@ -0,0 +1,4 @@
+{
+  "type": "minor",
+  "description": "Add LiteLLM chat and embedding model providers."
```
AlonsoGuevara, line 3: Is it intended that this commit includes both litellm and the token encoder change?

graphrag/language_model/providers/litellm/embedding_model.py
```diff
@@ -0,0 +1,274 @@
...
+    if model_config.api_version:
+        base_args["api_version"] = model_config.api_version
+    if model_config.api_key:
+        base_args["api_key"] = model_config.api_key
+    if model_config.organization:
+        base_args["organization"] = model_config.organization
+    if model_config.proxy:
+        base_args["proxy"] = model_config.proxy
+    if model_config.audience:
+        base_args["audience"] = model_config.audience
```
AlonsoGuevara: nit: This could be replaced with something like:

``` python
for f in ("api_base", "api_version", "api_key", "organization", "proxy", "audience"):
    v = getattr(model_config, f, None)
    if v is not None:
        base_args[f] = v
```

graphrag/language_model/providers/litellm/chat_model.py
```diff
@@ -0,0 +1,412 @@
...
+        self.name = name
+        self.config = config
+        self.cache = cache.child(self.name) if cache else None
+        self.completion, self.acompletion = _create_completions(
+            config, self.cache, "chat"
+        )
+
+    def _get_kwargs(self, **kwargs: Any) -> dict[str, Any]:
+        """Get model arguments supported by litellm."""
+        args_to_include = [
```
andresmor-ms, line 229: could change this for a set instead of a list :)

graphrag/language_model/providers/litellm/embedding_model.py
AlonsoGuevara: Or
```
fields = {"api_base", "api_version", "api_key", "organization", "proxy", "audience"}
base_args.update({k: v for k, v in vars(model_config).items() if k in fields and v is not None})
```

Both need testing tho, but will save you some lines of code

graphrag/language_model/providers/litellm/chat_model.py
```diff
@@ -0,0 +1,412 @@
...
+            "logprobs",
+            "top_logprobs",
+            "parallel_tool_calls",
+            "web_search_options",
+            "extra_headers",
+            "functions",
+            "function_call",
+            "thinking",
+        ]
+        new_args = {k: v for k, v in kwargs.items() if k in args_to_include}
```
andresmor-ms, line 250: if changed for a set you could also do something like:
```{k: kwargs[k] for k in args_to_include.intersection(kwargs)}```

File: graphrag/language_model/providers/litellm/embedding_model.py [hunk 7]
```diff
@@ -0,0 +1,274 @@
...
+        base_args["azure_ad_token_provider"] = get_bearer_token_provider(
+            DefaultAzureCredential(),
+            COGNITIVE_SERVICES_AUDIENCE,
+        )
+
+    def _base_embedding(**kwargs: Any) -> EmbeddingResponse:
+        new_args = {**base_args, **kwargs}
+
+        if "name" in new_args:
+            new_args.pop("name")
```
andresmor-ms, line 84: maybe use 
```del new_args["name"]``` since you are not doing anything with the result

File: graphrag/language_model/providers/litellm/services/rate_limiter/static_rate_limiter/__init__.py [hunk 8]
```diff
@@ -0,0 +1,124 @@
...
+from collections.abc import Iterator
+from contextlib import contextmanager
+from typing import Any
+
+from graphrag.language_model.providers.litellm.services.rate_limiter.rate_limiter import (
+    RateLimiter,
+)
+
+
+class StaticRateLimiter(RateLimiter):
```
AlonsoGuevara, line 18: Not a fan or large class definitions in the init.py
Can we mode it to its own file?

graphrag/language_model/providers/litellm/get_cache_key.py
```diff
@@ -0,0 +1,134 @@
...
+        "parameters": _get_parameters(model_config, **kwargs),
+    }
+
+    if "messages" in kwargs:
+        cache_key["messages"] = kwargs["messages"]
+    elif "input" in kwargs:
+        cache_key["input"] = kwargs["input"]
+    else:
+        msg = "Either 'messages' or 'input' must be provided in kwargs."
+        raise ValueError(msg)
```
andresmor-ms: why not move them as optional in the signature since one of them is required and add it in the docstring? also would be good to add docs about which one takes priority if both are defined (or throw an error if both are defined)

graphrag/language_model/providers/litellm/services/retry/exponential_retry/__init__.py
```diff
@@ -0,0 +1,75 @@
...
+import time
+from collections.abc import Awaitable, Callable
+from typing import Any
+
+from graphrag.language_model.providers.litellm.services.retry.retry import Retry
+
+logger = logging.getLogger(__name__)
+
+
+class ExponentialRetry(Retry):
```
AlonsoGuevara, line 18: For simplicity I would just move them to its own file within retry/
Or, if you still want to keep nesting I would do something like retry/strategy/ and put all there, to make them easier to discover

graphrag/query/context_builder/local_context.py
```diff
@@ -155,7 +155,7 @@ def build_covariates_context(
 def build_relationship_context(
     selected_entities: list[Entity],
     relationships: list[Relationship],
-    token_encoder: tiktoken.Encoding | None = None,
+    tokenizer: Tokenizer = get_tokenizer(),
```
andresmor-ms: it is usually bad practice to set a default as a function, this causes that once the function is loaded (the file imported or the function itself) it will set that tokenizer and not a default that gets created every time it runs, it could be fine for this function in particular, but wanted to point this out

graphrag/index/text_splitting/text_splitting.py
```diff
@@ -83,44 +83,18 @@ def split_text(self, text: str | list[str]) -> Iterable[str]:
...
-    _allowed_special: Literal["all"] | set[str]
-    _disallowed_special: Literal["all"] | Collection[str]
-
     def __init__(
         self,
-        encoding_name: str = defs.ENCODING_MODEL,
-        model_name: str | None = None,
-        allowed_special: Literal["all"] | set[str] | None = None,
-        disallowed_special: Literal["all"] | Collection[str] = "all",
+        tokenizer: Tokenizer = get_tokenizer(),
```
AlonsoGuevara: Defaults that invoke functions is a bad practice in Python. It is preferable to handle them by code.

graphrag/tokenizer/tokenizer.py
```diff
@@ -0,0 +1,40 @@
+# Copyright (c) 2024 Microsoft Corporation.
+# Licensed under the MIT License
+
+"""Tokenizer Abstract Base Class."""
+
+from abc import ABC, abstractmethod
+
+
+class Tokenizer(ABC):
```
andresmor-ms, line 9: there's no default behavior here what about a protocol instead?

graphrag/tokenizer/litellm_tokenizer.py
```diff
@@ -0,0 +1,47 @@
...
+# Licensed under the MIT License
+
+"""LiteLLM Tokenizer."""
+
+from litellm import decode, encode  # type: ignore
+
+from graphrag.tokenizer.tokenizer import Tokenizer
+
+
+class LitellmTokenizer(Tokenizer):
```
AlonsoGuevara, line 11: Protocols and not inheritance

.semversioner/next-release/minor-20250813214258257076.json
dworthen: This PR was built on top of the litellm branch/PR which has not been accepted yet which is why it is showing both changes but really this PR just has the encoding changes.

graphrag/language_model/providers/litellm/embedding_model.py
dworthen: I had that but our ruff formatting rules forbid using del so I changed to pop.

graphrag/index/text_splitting/text_splitting.py
natoverse: Can we just move this instantiation inside the init method?

graphrag/query/context_builder/local_context.py
dworthen: I wasn't able to update num_tokens in a backward compatible way to lean on the new Tokenizer interface. The utility is used in places where a Tokenizer is not used such as text chunking where the ChunkingConfig specifies its own encoding model instead of leaning on a defined LanguageModel definition. I am hoping to clean this up more in v3. Plus there are multiple of these num_token utils throughout the codebase. I would like to clean up and remove these duplicates part of v3 and started that work here. I think perhaps a good approach is to add this functionality to the Tokenizer interface itself which I have now done and added to this PR. Let me know if you think there is a better approach or disagree with this.

graphrag/tokenizer/tokenizer.py
dworthen: I have added some functionality to the ABC to help reduce some duplications of utilities in the codebase. The example is num_tokens functionality is now part of the base class. Let me know if you still prefer a protocol and that num_tokens or other common functionality should remain as separate utilities.

graphrag/tokenizer/litellm_tokenizer.py
dworthen: I have added some functionality to the ABC to help reduce some duplications of utilities in the codebase. The example is num_tokens functionality is now part of the base class. Let me know if you still prefer a protocol and that num_tokens or other common functionality should remain as separate utilities.

graphrag/language_model/providers/litellm/embedding_model.py
dworthen: I removed the checks and simplified the base_args assignment.

graphrag/language_model/providers/litellm/services/rate_limiter/static_rate_limiter/__init__.py
dworthen: I reorganized this as suggested.

graphrag/language_model/providers/litellm/services/retry/exponential_retry/__init__.py
dworthen: I reorganized this as suggested.

graphrag/query/context_builder/local_context.py
dworthen: Moved instantiation/fallback into the method.

graphrag/index/text_splitting/text_splitting.py
dworthen: Moved instantiation/fallback into the method.

graphrag/config/defaults.py
```diff
@@ -23,6 +24,25 @@
...
+from graphrag.language_model.providers.litellm.services.retry.incremental_wait_retry import (
+    IncrementalWaitRetry,
+)
+from graphrag.language_model.providers.litellm.services.retry.native_wait_retry import (
+    NativeRetry,
+)
+from graphrag.language_model.providers.litellm.services.retry.random_wait_retry import (
+    RandomWaitRetry,
+)
+from graphrag.language_model.providers.litellm.services.retry.retry import Retry
```
AlonsoGuevara, line 45: You can export all these in the module's init. So you just have imports like: 
`from graphrag.language_model.providers.litellm.services.retry import ( Retry, RateLimiter, StaticRateLimiter, ExponentialRetry, IncrementalWaitRetry, NativeRetry, RandomWaitRetry) `

graphrag/config/models/graph_rag_config.py
```diff
@@ -89,6 +95,47 @@ def _validate_models(self) -> None:
...
+                rpm = (
+                    model.requests_per_minute
+                    if type(model.requests_per_minute) is int
+                    else None
+                )
+                tpm = (
+                    model.tokens_per_minute
+                    if type(model.tokens_per_minute) is int
+                    else None
+                )
```
andresmor-ms, line 133: just curious, what is model.tokens_per_minute type? can it be something else than int or None?

dworthen: ya fnllm supports the string "auto". I think these checks can go away in v3 when we pull out fnllm and can clean up some of the config field types.

PR #2055, opened 2025-09-15
feat: Make community summarization rate limit configurable

### LINKED ISSUE #2020: [Bug]: summarize communities RateLimiter

### Describe the bug

Whether it is Standard or Fast, the two versions of graphrag have severe rate limiting during the summarize stage. The token bucket algorithm should be used, but the rate limiting is once every 60 seconds and it is hard-coded. This leads to a very slow index stage, especially when there are many communities

```python
async def _run_extractor(
    model: ChatModel,
    community: str | int,
    input: str,
    level: int,
    args: StrategyConfig,
    callbacks: WorkflowCallbacks,
) -> CommunityReport | None:
    # RateLimiter
    rate_limiter = RateLimiter(rate=1, per=60)
    ...
    try:
        await rate_limiter.acquire()
        ...
```

### Steps to reproduce

_No response_

### Expected Behavior

_No response_

### GraphRAG Config Used

```yaml
# Paste your config here

```

### Logs and screenshots

_No response_

### Additional Information

- GraphRAG Version:v2.3.0 and v2.4.0
- Operating System:wsl
- Python Version:3.11
- Related Issues:

devampatel03:
## Description

Currently , the community summarization step is hard-coded to run only one request every 60 seconds. This makes the indexing process extremely slow if you have a lot of communities to summarize.

This PR fixes this by letting users configure the rate limit themselves in the YAML config, so they can speed it up to match what their LLM's API can handle.

## Related Issues

Fixes #2020

## Proposed Changes

- In `graphrag/index/operations/summarize_communities/strategies.py`, the `rate` and `per` values for the `RateLimiter` are now pulled from the strategy `args`.
- The old hard-coded values (`rate=1, per=60`) are now used as defaults, so existing configs won't break.

## Checklist

## Additional Notes

To use the new settings, you can add `rate` and `per` to your config file under a strategy that uses it. Here's an example for the summarization step:

```yaml
summarize_communities:
  strategy:
    type: graph_intelligence
    rate: 30 # e.g., 30 requests
    per: 60  # e.g., per 60 seconds

### DIFF
--- a/graphrag/index/operations/summarize_communities/strategies.py
+++ b/graphrag/index/operations/summarize_communities/strategies.py
@@ -52,7 +52,10 @@ async def _run_extractor(
     args: StrategyConfig,
 ) -> CommunityReport | None:
     # RateLimiter
-    rate_limiter = RateLimiter(rate=1, per=60)
+    # Read rate and per from the config, defaulting to 1 per 60 if not provided
+    rate = args.get("rate", 1)
+    per = args.get("per", 60)
+    rate_limiter = RateLimiter(rate=rate, per=per)
     extractor = CommunityReportsExtractor(
         model,
         extraction_prompt=args.get("extraction_prompt", None),


natoverse:
Hi @devampatel03, thanks for your contribution!

Please take a look at this PR that I just pushed: https://github.com/microsoft/graphrag/pull/2056

We have actually had a way to configure this for some time, but the hard-coded RateLimiter was blocking it.

devampatel03:
Makes perfect sense, @natoverse . Appreciate you explaining the cleaner approach.

PR #2056, opened 2025-09-15
Remove community reports rate limiter

### LINKED ISSUE #2020: [Bug]: summarize communities RateLimiter

natoverse:
This removes the hard-coded rate limited built into community report summarization. This was added over a year ago to address throttling issues, but the rate limit was hard-coded to 1 RPM. Newer versions of GraphRAG allow specification of independent model configs per workflow step, so this can be handled entirely via config now.

For example (truncated settings.yaml):

```yaml
models:
  default_chat_model:
    ...
    requests_per_minute: 500
  community_chat_model:
    ...
    requests_per_minute: 1

# later...
community_reports:
    model_id: community_chat_model
    ...
```

In this example, the default chat model could run faster, with a separate config for the community chat model to control a slower rate.

Closes #2020


graphrag/cache/factory.py
```diff
@@ -97,8 +97,13 @@ def create_cosmosdb_cache(**kwargs) -> PipelineCache:
...
 
+def create_noop_cache(**_kwargs) -> PipelineCache:
+    """Create a no-op cache implementation."""
+    return NoopPipelineCache()
+
+
 # --- register built-in cache implementations ---
-CacheFactory.register(CacheType.none.value, NoopPipelineCache)
+CacheFactory.register(CacheType.none.value, create_noop_cache)
 CacheFactory.register(CacheType.memory.value, InMemoryCache)
```
AlonsoGuevara: Just by the look of the other fields, shouldn't this line be a "creator" also?

natoverse: NoopPipelineCache init takes no args, so it fails on --no-cache. The factory function fixes that (was a regression from the copilot factory PR)

AlonsoGuevara: @natoverse  but I meant the InMemoryCache. I see all the other factory registers use a creator method. Shouldn't InMemory do the same?

natoverse: I see. I have fixed this now for consistency. I think it was not using a factory because the constructor itself was functional as a constructor, but I like the clarity

AlonsoGuevara (approved):

PR #2083, opened 2025-10-01
Litellm auth fix

natoverse:
This fixes an auth issue with LiteLLm when providing custom scope. LiteLLM does not like the `audience` parameter our config uses, instead using `azure_scope`.

I also reworded `max_attempts` as `max_retries` to reflect the logic, and reworked the model config connectivity test to loop through all dict items and check any from our known type set.


natoverse:
Note that this PR took some updating of the smoke test config to use newer model (4.1) and parameters around that. While setting that I ended up cleaning out a bunch of old references to env vars that either weren't used anymore or where in the GitHub secrets but aren't actually "secrets"

dworthen (approved):

PR #2086, opened 2025-10-06
Housekeeping

natoverse:
Little bit of cleanup leading into 2.7
- fixed a few places where a token encoder was still being used instead of tokenizer
- fixed the example notebooks
- fixed the dev deps group
- added deprecation warnings for multi search and fnllm, leading to removal in 3.0


graphrag/query/question_gen/local_gen.py
```diff
@@ -118,7 +117,7 @@ async def agenerate(
                 },
                 completion_time=time.time() - start_time,
                 llm_calls=1,
-                prompt_tokens=num_tokens(system_prompt, self.token_encoder),
+                prompt_tokens=num_tokens(system_prompt, self.tokenizer),
```
andresmor-ms: I think this should be `self.tokenizer.num_tokens(system_prompt)` now

graphrag/query/llm/text_utils.py
```diff
@@ -9,19 +9,20 @@
...
 
 import graphrag.config.defaults as defs
+from graphrag.tokenizer.get_tokenizer import get_tokenizer
+from graphrag.tokenizer.tokenizer import Tokenizer
 
 logger = logging.getLogger(__name__)
 
 
-def num_tokens(text: str, token_encoder: tiktoken.Encoding | None = None) -> int:
+def num_tokens(text: str, tokenizer: Tokenizer | None = None) -> int:
```
andresmor-ms: Do we want to keep this? the Tokenizer implements this too

andresmor-ms (approved):

PR #2105, opened 2025-10-17
Unified factory

natoverse:
Consolidates the factories around a common base class.

### DIFF
--- a/graphrag/language_model/factory.py
+++ b/graphrag/language_model/factory.py
@@ -21,8 +21,5 @@ class EmbeddingModelFactory(Factory[EmbeddingModel]):
 
 
 # --- Register default implementations ---
-ChatModelFactory().register(ModelType.Chat, lambda **kwargs: LitellmChatModel(**kwargs))
-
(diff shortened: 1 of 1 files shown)


graphrag/language_model/factory.py
```diff
@@ -3,101 +3,26 @@
...
-            model_type
-        ) or cls.is_supported_embedding_model(model_type)
+class EmbeddingModelFactory(Factory[EmbeddingModel]):
+    """Singleton factory for creating EmbeddingModel instances."""
 
 
 # --- Register default implementations ---
-ModelFactory.register_chat(ModelType.Chat, lambda **kwargs: LitellmChatModel(**kwargs))
-ModelFactory.register_embedding(
+ChatModelFactory().register(ModelType.Chat, lambda **kwargs: LitellmChatModel(**kwargs))
```
dworthen: Are the lambdas necessary here? Is it possible/desirable to just use the class constructor?

natoverse: Yep, those look like leftovers

PR #2106, opened 2025-10-18
Prefix vector store

gaudyb:
## Description

Add a new property to the vector store configuration: prefix to add as a prefix for all the embedding indexes.

## Checklist

### DIFF
--- a/graphrag/config/init_content.py
+++ b/graphrag/config/init_content.py
@@ -77,7 +77,6 @@ reporting:
 vector_store:
   type: {vector_store_defaults.type}
   db_uri: {vector_store_defaults.db_uri}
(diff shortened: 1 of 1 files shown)


graphrag/config/init_content.py
```diff
@@ -77,7 +77,7 @@
 vector_store:
   type: {vector_store_defaults.type}
   db_uri: {vector_store_defaults.db_uri}
-  container_name: {vector_store_defaults.container_name}
+  index_prefix: {vector_store_defaults.index_prefix}
```
natoverse: Can you remove this from `init_content` completely? I'd like it to be a lesser-used option, so keeping it out of the default settings.yaml will make it less obvious.

Issue #2115, opened 2025-10-24
[Bug]: Blob storage loads all files regardless of `base_dir`


Robotuks:
### Do you need to file an issue?

- [x] I have searched the existing issues and this bug is not already filed.
- [x] My model is hosted on OpenAI or Azure. If not, please look at the "model providers" issue and don't file a new one here.
- [x] I believe this is a legitimate bug, not just a question. If this is a question, please use the Discussions area.

### Describe the bug

Using blob storage for input. But when running indexing noticed that it tries to read more files than it should. It gives me all files within a container.

input config example: 
```
input:
  storage:
    type: blob # [file, blob]
    base_dir: "input/folder_1/"
    container_name: "graphrag-container"
    storage_account_blob_url: "https://graphragexample.blob.core.windows.net"
  file_type: text
```

Using this example config, getting warning messages during indexing like: 
`2025-10-24 11:30:38.0647 - WARNING - graphrag.storage.blob_pipeline_storage - Error getting key input/folder_1/input/folder_2/text.txt`

In code, noticed you are getting all blobs from the container:
https://github.com/microsoft/graphrag/blob/fdb7e3835badc483f50605db476d8636aeea7156/graphrag/storage/blob_pipeline_storage.py#L146

Couldn't pin point exact place why other directories are not being filtered out but you could just get files under a base dir to avoid this issue:
`all_blobs = list(container_client.list_blobs(name_starts_with=base_dir))`

### Additional Information

- GraphRAG Version: 2.6.0
- Operating System: Ubuntu 22
- Python Version: 3.12

Robotuks:
Did some testing. Noticed that `base_dir` is not passed to `.find()` method [here](https://github.com/microsoft/graphrag/blob/fdb7e3835badc483f50605db476d8636aeea7156/graphrag/index/input/util.py#L26)

By adding `base_dir` argument fixed my issue:
```
    files = list(
        storage.find(
            re.compile(config.file_pattern),
            base_dir=config.storage.base_dir,
            file_filter=config.file_filter,
        )
    )
```

natoverse:
This is resolved on our v3 branch, with expected release to main and pypi in the next couple of weeks

PR #2118, opened 2025-11-04
Storage fixes and cleanup

natoverse:
- Fixes recursion and filtering issue in blobs
- Simplifies API by removing unused params
  - Removes base_dir - this was not used and should use what is sent during init anyway
  - Removes max_count - this was never used, and should use a proper paging mechanism if we want it
- Cleans up config defaults and validation

### DIFF
--- a/packages/graphrag/graphrag/config/defaults.py
+++ b/packages/graphrag/graphrag/config/defaults.py
@@ -24,7 +24,9 @@ from graphrag.index.operations.build_noun_graph.np_extractors.stop_words import 
     EN_STOP_WORDS,
 )
 
+DEFAULT_INPUT_BASE_DIR = "input"
 DEFAULT_OUTPUT_BASE_DIR = "output"
+DEFAULT_UPDATE_OUTPUT_BASE_DIR = "update_output"
 DEFAULT_CHAT_MODEL_ID = "default_chat_model"
 DEFAULT_CHAT_MODEL_TYPE = ModelType.Chat
 DEFAULT_CHAT_MODEL_AUTH_TYPE = AuthType.APIKey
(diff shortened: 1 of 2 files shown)


natoverse:
Fixes #2115

packages/graphrag/graphrag/config/defaults.py
```diff
@@ -240,7 +240,7 @@ class StorageDefaults:
 class InputStorageDefaults(StorageDefaults):
     """Default values for input storage."""
 
-    base_dir: str = "input"
+    base_dir: str | None = "input"
```
gaudyb: Isn't better to add the value "input" as a DEFAULT_BASE_DIR value?

packages/graphrag/graphrag/config/defaults.py
```diff
@@ -362,7 +362,7 @@ class SummarizeDescriptionsDefaults:
 class UpdateIndexOutputDefaults(StorageDefaults):
     """Default values for update index output."""
 
-    base_dir: str = "update_output"
+    base_dir: str | None = "update_output"
```
gaudyb: Same as above

packages/graphrag/graphrag/config/models/graph_rag_config.py
```diff
@@ -167,14 +167,16 @@ def _validate_input_base_dir(self) -> None:
 
     output: StorageConfig = Field(
         description="The output configuration.",
-        default=StorageConfig(),
+        default=StorageConfig(
+            base_dir=graphrag_config_defaults.output.base_dir,
```
dworthen, line 171: Curious why we don't do `StorageConfig(**asdict(graphrag_config_defaults.output))` here and in all other places we instantiate an object instead of plucking which default values we want. 

Don't think we need to change anything for this PR as this pattern has been around but maybe at some point in the future, we may want to revisit how we map defaults into the application. I find it interesting that we have a defaults hierarchy object but don't use it as such but instead we pluck from it like it is a list of consts.

natoverse: It's worth reviewing. For these in particular, we have three different storage instances that only vary in their base_dir

PR #2120, opened 2025-11-05
Nov 2025 housekeeping

natoverse:
- splits CI to allow separate static checks and unit tests
- removed gensim sideload
- bumps core matrix to python 3.11 and 3.12, omitting 3.10
- full sweep through docs to ensure accuracy with latest v3 changes


dworthen:
You missed updating a python version in `github/workflows/python-publish.yml`.

PR #2124, opened 2025-11-10
Fix type mismatch in DynamicCommunitySelection children lookup

### LINKED ISSUE #2004: [Bug]: `DynamicCommunitySelection` skips child communities because `children` IDs are int while `self.reports` keys are str

### Describe the bug

In graphrag/query/context_builder/dynamic_community_selection.py, lines 131-132 read:

```
for child in self.communities[community].children:  # child is int
    if child in self.reports:                       # self.reports keys are str
```
Because children holds integers and self.reports is keyed by strings, the membership test always fails.
As a result, IDs that should be selected as child communities are missed, and the exploration of the next community level is never executed.

### Steps to reproduce

Execute `global_search` with `dynamic_community_selection=True` using a knowledge graph where the highest `community_level` is 1 or higher.

### Expected Behavior

**Expected Behavior**
Child community IDs should match the type used in `self.reports`, allowing them to be evaluated and, if relevant, included in the result.

**Actual Behavior**
`if child in self.reports` is always `False`; all child communities are skipped, so the algorithm never descends to the next level.

### GraphRAG Config Used

```yaml
# Paste your config here

```

### Logs and screenshots

This is the current behavior when a query is issued:
The message cannot find community is displayed, and only 10 out of 5,586 communities are referenced.
Image

This is the behavior after a simple fix was applied:
309 out of 5,586 communities are now referenced when a query is issued.
Note: The `threshold` is set to 2.
Image

Fixed Code:
```
for child in self.communities[community].children: 
   if str(child) in self.reports: 
      communities_to_rate.append(str(child))
```

### Additional Information

- GraphRAG Version: v2.4.0
- Python Version: 3.12.11
- Related Issues: Nothing

KeonhoChu:
## Description

  Fixes a type mismatch bug in `DynamicCommunitySelection` that prevented the algorithm from correctly traversing child communities during global search queries with `dynamic_community_selection=True`.

  ## Related Issues

  Fixes #2004

  ## Proposed Changes

  - **Fixed type mismatch in `dynamic_community_selection.py` (lines 125-129)**
    - Child community IDs from `self.communities[community].children` were not matching keys in `self.reports` dictionary due to type inconsistency
    - Added explicit string conversion (`str(child)`) to ensure child IDs match the `self.reports` key type
    - Added explanatory comment for clarity

  **Root Cause:**
  - `Community.children` contains IDs that may be loaded as integers
  - `CommunityReport.community_id` is defined as `str` type
  - `self.reports` dictionary is keyed by `community_id` (strings)
  - The membership test `if child in self.reports` always failed, causing all child communities to be skipped

  **Impact:**
  - Before fix: Only 10 out of 5,586 communities were referenced (stuck at level 0)
  - After fix: 309 out of 5,586 communities are correctly traversed across multiple levels

  ## Checklist

  - [v] I have tested these changes locally (unit tests pass)
  - [v] I have reviewed the code changes
  - [v] I have updated the documentation (added inline comment explaining the fix)
  - [v] I have added a semversioner changeset (patch level)

### DIFF
--- a/.semversioner/next-release/patch-20251110004953346654.json
+++ b/.semversioner/next-release/patch-20251110004953346654.json
@@ -1,0 +1,4 @@
+{
+  "type": "patch",
+  "description": "Fix type mismatch in DynamicCommunitySelection when checking children communities. Convert child IDs to string to match self.reports key type (fixes #2004)"
+}
--- a/graphrag/query/context_builder/dynamic_community_selection.py
+++ b/graphrag/query/context_builder/dynamic_community_selection.py
@@ -123,8 +123,10 @@ class DynamicCommunitySelection:
                     # TODO check why some sub_communities are NOT in report_df
                     if community in self.communities:
                         for child in self.communities[community].children:
-                            if child in self.reports:
-                                communities_to_rate.append(child)
+                            # Convert child to string to match self.reports key type
+                            child_str = str(child)
+                            if child_str in self.reports:
+                                communities_to_rate.append(child_str)
                             else:
                                 logger.debug(
                                     "dynamic community selection: cannot find community %s in reports",


Keonho-Chu:

Keonho-Chu:
Thanks for addressing this in #2159! I'm glad my idea in this PR helped resolve the issue. I'll close this now.

PR #2127, opened 2025-11-11
Add graphrag-storage.

dworthen:
Add graphrag-storage package.

### DIFF
--- a/packages/graphrag-storage/graphrag_storage/storage_factory.py
+++ b/packages/graphrag-storage/graphrag_storage/storage_factory.py
@@ -6,7 +6,7 @@
 
 from collections.abc import Callable
 
-from graphrag_common.factory import Factory
+from graphrag_common.factory import Factory, ServiceScope
 
 from graphrag_storage.storage import Storage
 from graphrag_storage.storage_config import StorageConfig
@@ -21,7 +21,9 @@ storage_factory = StorageFactory()
 
 
 def register_storage(
-    storage_type: str, storage_initializer: Callable[..., Storage]
+    storage_type: str,
+    storage_initializer: Callable[..., Storage],
+    scope: ServiceScope = "transient",
 ) -> None:
     """Register a custom storage implementation.
 
@@ -32,7 +34,7 @@ def register_storage(
         - storage_initializer: Callable[..., Storage]
             The storage initializer to register.
     """
-    storage_factory.register(storage_type, storage_initializer)
+    storage_factory.register(storage_type, storage_initializer, scope)
 
 
 def create_storage(config: StorageConfig) -> Storage:
--- a/packages/graphrag-common/graphrag_common/factory/__init__.py
+++ b/packages/graphrag-common/graphrag_common/factory/__init__.py
@@ -3,6 +3,6 @@
 
 """The GraphRAG factory module."""
 
-from graphrag_common.factory.factory import Factory
+from graphrag_common.factory.factory import Factory, ServiceScope
 
-__all__ = ["Factory"]
+__all__ = ["Factory", "ServiceScope"]


packages/graphrag-storage/pyproject.toml
```diff
@@ -0,0 +1,47 @@
...
+    {name = "Kate Lytvynets", email = "kalytv@microsoft.com"},
+    {name = "Kenny Zhang", email = "zhangken@microsoft.com"},
+    {name = "Mónica Carvajal"},
+    {name = "Nathan Evans", email = "naevans@microsoft.com"},
+    {name = "Rodrigo Racanicci", email = "rracanicci@microsoft.com"},
+    {name = "Sarah Smith", email = "smithsarah@microsoft.com"},
+]
+license = "MIT"
+readme = "README.md"
+license-files = ["LICENSE"]
```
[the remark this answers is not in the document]
dworthen: Its dynamically included at build time.

docs/config/yaml.md
```diff
@@ -81,7 +81,7 @@ Our pipeline can ingest .csv, .txt, or .json data from an input location. See th
 #### Fields
 
 - `storage` **StorageConfig**
-  - `type` **file|blob|cosmosdb** - The storage type to use. Default=`file`
+  - `type` **FileStorage|AzureBlobStorage|AzureCosmosStorage** - The storage type to use. Default=`FileStorage`
```
natoverse: I'm iffy on this change. It breaks with current convention, and feels "off" for how you would put values in yaml. I understand the intent from factory registration perspective of course. Maybe worth tagging a couple of other folks to see if there is consensus one way or the other

packages/graphrag-storage/graphrag_storage/azure_blob_storage.py
```diff
@@ -12,53 +12,65 @@
...
 
-    def __init__(self, **kwargs: Any) -> None:
+    def __init__(
+        self,
+        base_dir: str | None = None,
+        connection_string: str | None = None,
+        storage_account_blob_url: str | None = None,
+        container_name: str | None = None,
+        encoding: str = "utf-8",
+        **kwargs: Any,
```
natoverse, line 41: Presumably the kwargs in these constructors are just to soak additional unknown args from the factory? My preference is definitely to avoid kwargs.get and instead explicitly name it – would it therefore make sense register lambdas with the factory that pluck exact args and perform this soaking function? It's a bit more code, but could set a reasonable pattern of explicit mapping from the factory init call to instances?

packages/graphrag-storage/graphrag_storage/azure_blob_storage.py
```diff
@@ -245,7 +257,7 @@ async def get_creation_date(self, key: str) -> str:
             return ""
 
 
-def validate_blob_container_name(container_name: str):
+def _validate_blob_container_name(container_name: str) -> None:
```
natoverse, line 255: Could all of these checks be collapsed into a single regex?

packages/graphrag-storage/graphrag_storage/file_storage.py
```diff
@@ -16,26 +16,33 @@
...
         """Create a file based storage."""
-        self._base_dir = kwargs.get("base_dir", "")
-        self._encoding = kwargs.get("encoding", "utf-8")
+        if base_dir is None:
+            msg = "FileStorage requires a base_dir to be specified."
+            logger.error(msg)
+            raise ValueError(msg)
+
+        self._base_dir = Path(base_dir).resolve()
+        self._encoding = encoding
```
natoverse, line 36: Given that the get/set include optional encoding, should we take it off the init kwargs? Or maybe the reverse - since it isn't applicable to all storage types, maybe we take it off the get/set and expect any given storage instance to only work with a single encoding via init args?

packages/graphrag-storage/graphrag_storage/storage_config.py
```diff
@@ -0,0 +1,41 @@
...
+
+"""Storage configuration model."""
+
+from pydantic import BaseModel, ConfigDict, Field
+
+
+class StorageConfig(BaseModel):
+    """The default configuration section for storage."""
+
+    model_config = ConfigDict(extra="allow")
```
natoverse, line 14: This doesn't seem like the right name for this, and maybe should be punted to a broader discussion about how we want to handle open-ended config

packages/graphrag-storage/graphrag_storage/azure_blob_storage.py
dworthen: Done. Collapsed all the checks into a single regex.

docs/config/yaml.md
dworthen: I switched to using string enums that are more aligned with yaml usage.

packages/graphrag-storage/graphrag_storage/file_storage.py
dworthen: I added encoding to init args through StorageConfig and set implementation specific defaults.

packages/graphrag-storage/graphrag_storage/azure_blob_storage.py
dworthen: I removed kwargs in these inits to avoid swallowing unknown config parameters. Now if you try to pass unsupported params to an implementation it will throw an exception. I prefer this approach over lambdas as now it is the default behavior and all impls get this validation so users creating custom impls don't also have to create custom lambdas to get the same kind of validation and plucking behavior. Let me know what you think of this approach.

packages/graphrag-storage/graphrag_storage/storage_config.py
dworthen: Not sure what else to call it. Got any suggestions?

natoverse: We could go a couple of ways: (1) create a bespoke extra config name for each provider config type to hold these. In this case maybe `storage_config`, or (2) re-use a generic name across all providers, such as `provider_config` or `extra_config` or some such.

packages/graphrag-storage/graphrag_storage/azure_blob_storage.py
natoverse: The way I read this functionality currently: (1) the old kwargs has been expanded into named params for clarity, which I agree with. Presumably something upstream is sending these as kwargs since they may not all be allowed for varying providers and this is unknown to the config system, (2) the remaining kwargs at the end of this function signature is to soak those unknowns and ignore them.

packages/graphrag-storage/graphrag_storage/storage_factory.py
```diff
@@ -0,0 +1,81 @@
...
+    """
+    config_model = config.model_dump()
+    storage_strategy = config_model.pop("type")
+
+    # Check storage_strategy is a string
+    if not isinstance(storage_strategy, str):
+        msg = f"StorageConfig.type must be a string, got {type(storage_strategy)}"
+        raise TypeError(msg)
+
+    if storage_strategy not in storage_factory:
```
natoverse, line 56: Seems like we could simplify this with the creator registration. One way to think of it would be: all users of the factory can register a provider impl, we just pre-register a set of supported ones. So that pre-registration should look exactly like how someone else might do it, with the exception that we might do a dynamic import whereas custom stuff wouldn't need it.

For example:
```
def register_file():
    from graphrag_storage.file_storage import FileStorage
    return FileStorage
    

register_storage(StorageType.FILE, register_file)
```
This would elimiate the match case. It's a bit like the lambdas that were sometimes used before, which may seem redundant, but would allow us to avoid hard-coding a bunch of providers (you could imagine these being register elsewhere, for example)

packages/graphrag-storage/graphrag_storage/storage_type.py
```diff
@@ -0,0 +1,16 @@
...
+
+"""Builtin storage implementation types."""
+
+from enum import StrEnum
+
+
+class StorageType(StrEnum):
+    """Enum for storage types."""
+
+    FILE = "file"
```
natoverse: I think we've been generally moving to TitleCase, which would match cross-platform stuff a little more too.

File
Memory
AzureBlob
AzureCosmos

packages/graphrag/graphrag/cache/factory.py
```diff
@@ -30,19 +30,22 @@ class CacheFactory(Factory[PipelineCache]):
 # --- register built-in cache implementations ---
 def create_file_cache(**kwargs) -> PipelineCache:
```
natoverse, line 28: These would need to dynamically import their storage in the creator functions too

packages/graphrag/graphrag/config/defaults.py
```diff
@@ -230,12 +231,13 @@ class GlobalSearchDefaults:
...
 
-    type: ClassVar[StorageType] = StorageType.file
+    type: str = StorageType.FILE
+    encoding: str | None = None
     base_dir: str | None = None
-    connection_string: None = None
-    container_name: None = None
-    storage_account_blob_url: None = None
-    cosmosdb_account_url: None = None
+    azure_connection_string: None = None
```
natoverse, line 237: I think if we want to rename these we may want to go the opposite direction and have a handful of generic and generally obvious params that are common to lots of impls.

connection_string
container_name
account_url (to replace both blob and cosmos account urls)

thoughts?

natoverse: Another option (maybe a bigger discussion) would be to take the approach we are moving to with the LM config, which would be to have something like an `init_args` or similar that is just a dict, and completely custom per provider type.

packages/graphrag-storage/graphrag_storage/storage_type.py
```diff
@@ -0,0 +1,16 @@
...
+    MEMORY = "memory"
+    AZURE_BLOB = "azure_blob"
```
natoverse: I don't see a super strong reason to change these, given that in our world "blob" and "cosmos" both mean azure. Just more typing for a level of flexibility that probably isn't important...

packages/graphrag-storage/graphrag_storage/azure_blob_storage.py
dworthen: That was correct but I removed kwargs altogether so point (2) is no longer correct. The above code is outdated. Implementations need to list out init args and any additional args provided to an implementation will throw an exception.

packages/graphrag-storage/graphrag_storage/storage_config.py
dworthen: I am now realizing your issue is with `model_config` and not `StorageConfig`. `model_config` is a special keyword for pydantic and is the way you configure pydantic and is not actually exposed to the end user. `model_config = ConfigDict(extra="allow")` instructs pydantic that the below fields are expected but users may also provide additional fields and to not throw errors for additional fields. This allows end users to add arbitrary settings if using a custom registered implementation. Check out the readme for an example.

packages/graphrag-storage/graphrag_storage/storage_factory.py
dworthen: I am not sure how you get around not hard-coding a bunch of providers. Even when you have these lambdas you end up with 

```python
register_storage(StorageType.FILE, _create_file_storage)
register_storage(StorageType.MEMORY, _create_memory_storage)
register_storage(StorageType.AZURE_BLOB, _create_azure_blob_storage)
register_storage(StorageType.AZURE_COSMOS, _create_azure_cosmos_storage)
```

and need to manage adding and removing built-in providers. Could try to iterate over the enum but then you need some sort of dictionary mapping of enums to import paths and modules which I think is needlessly complex for cleverness. 

I don't mind the lambda approach, solves the same thing as the match case. One aspect that is nice about the match case is the logic of registering is within a function as opposed to a top-level module register and I do prefer trying to avoid code running at module import time. Thoughts?

packages/graphrag-storage/graphrag_storage/storage_type.py
dworthen: Updated to use TitleCase property names.

packages/graphrag/graphrag/cache/factory.py
dworthen: Done. Cache factory now dynamically imports underlying storage providers.

packages/graphrag-storage/graphrag_storage/storage_type.py
dworthen: I reverted the values back to what they were in GR. We can do a sweep later if we want to rename the values.

packages/graphrag-storage/graphrag_storage/storage_config.py
```diff
@@ -0,0 +1,48 @@
...
+
+    azure_container_name: str | None = Field(
+        description="The Azure Blob Storage container name or CosmosDB container name to use.",
+        default=None,
+    )
+    azure_account_url: str | None = Field(
+        description="The account url for Azure Blob Storage or Azure CosmosDB.",
+        default=None,
+    )
+    azure_cosmosdb_database_name: str | None = Field(
```
natoverse: Are we going to remove some of the tech specifics here? (remove azure*, cosmos*) so these are more generic

connection_string
container_name
account_url
database_name

packages/graphrag-storage/graphrag_storage/storage_factory.py
natoverse: I think we decided to pull this to registrations so it provides a clear example for third parties?

dworthen: We decided to keep the registration dynamic in the `create_storage` method so that 

1. Users can import the raw `storage_factory` factory and use it without having any prior registrations. That way it acts as a library in the sense it does not make any assumptions about use and does not preregister any impls. 
2. The `create_storage` method provides a happy path that works with concrete, well-typed configs instead of dicts and will preregister our common, built-in impls. That way we reduce boilerplate and support our common use cases out-of-the-box.

natoverse: ok, right

PR #2158, opened 2025-12-28
Fix DynamicCommunitySelection type mismatch for children IDs

### LINKED ISSUE #2004: [Bug]: `DynamicCommunitySelection` skips child communities because `children` IDs are int while `self.reports` keys are str

majiayu000:
## Summary
- Fix type mismatch in `DynamicCommunitySelection.select()` where children IDs (integers) were not matching string report keys
- Convert `child` to string before checking membership in `self.reports`

## Problem
In `dynamic_community_selection.py`, the children list may contain integers while `self.reports` keys are strings. This causes `if child in self.reports` to always return `False`, skipping all child communities during dynamic selection.

## Solution
Convert children IDs to strings using `str(child)` before checking membership and appending to `communities_to_rate`.

## Test Plan
- Added unit tests for both integer and string children scenarios
- All existing unit tests pass
- Static checks (ruff, pyright) pass

Fixes #2004

### DIFF
--- a/.semversioner/next-release/patch-20251228023802955549.json
+++ b/.semversioner/next-release/patch-20251228023802955549.json
@@ -1,0 +1,4 @@
+{
+  "type": "patch",
+  "description": "Fix DynamicCommunitySelection type mismatch where children IDs as integers were not matching string report keys"
+}
--- a/graphrag/query/context_builder/dynamic_community_selection.py
+++ b/graphrag/query/context_builder/dynamic_community_selection.py
@@ -123,12 +123,14 @@ class DynamicCommunitySelection:
                     # TODO check why some sub_communities are NOT in report_df
                     if community in self.communities:
                         for child in self.communities[community].children:
-                            if child in self.reports:
-                                communities_to_rate.append(child)
+                            # Ensure child is a string to match self.reports keys
+                            child_id = str(child)
+                            if child_id in self.reports:
+                                communities_to_rate.append(child_id)
                             else:
                                 logger.debug(
                                     "dynamic community selection: cannot find community %s in reports",
-                                    child,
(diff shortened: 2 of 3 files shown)


gaudyb:
Implemented in PR #2159 using this logic and #2124 PRs. Thank you!

PR #2172, opened 2026-01-14
Vector package

natoverse:
Extracts the vector stores to a monorepo package and cleans up the config

### DIFF
--- a/packages/graphrag-vectors/graphrag_vectors/azure_ai_search.py
+++ b/packages/graphrag-vectors/graphrag_vectors/azure_ai_search.py
@@ -22,7 +22,6 @@ from azure.search.documents.indexes.models import (
 )
 from azure.search.documents.models import VectorizedQuery
 
-from graphrag_vectors.types import TextEmbedder
 from graphrag_vectors.vector_store import (
     VectorStore,
     VectorStoreDocument,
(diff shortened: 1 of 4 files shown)


packages/graphrag-vectors/graphrag_vectors/__init__.py
```diff
@@ -0,0 +1,40 @@
...
+from graphrag_vectors.vector_store_config import VectorStoreConfig
+from graphrag_vectors.vector_store_factory import (
+    VectorStoreFactory,
+    create_vector_store,
+    register_vector_store,
+    vector_store_factory,
+)
+from graphrag_vectors.vector_store_type import VectorStoreType
+
+__all__ = [
```
dworthen, line 22: I don't tend to expose impls in the init and instead expose the types, create, and register factory methods. That way when people `from graphrag_vectors import create_vector_store` directly from the package module they still get the benefit of the impls being lazy loaded.

natoverse: Ah, right

packages/graphrag/graphrag/config/models/graph_rag_config.py
```diff
@@ -276,13 +277,26 @@ def _validate_reporting_base_dir(self) -> None:
     )
     """The basic search configuration."""
 
+    def _validate_vector_store(self) -> None:
```
dworthen, line 280: I find it confusing that we validate vector store and the vector store uri both in graphrag and within the vector store package. I think as we pull out the packages and config that all the validation should go in the package whether through model validation or impl validation.

natoverse: Yeah, I had the same concerns as I was restructuring this. The issue is that there are some config validations that are agnostic to the core package, and there are some that are specific to the application. In this case, all of the embeddings schema stuff is specific to GraphRAG's utilization of the package in a way that ensures all required config for a specific set of embeddings is present, but those particular embeddings are not at all relevant to the core package. I think one way to resolve this could be to remove the validation in the package config entirely and shift it to the impls (the primary one here being the URLs).

PR #2174, opened 2026-01-14
Add GraphRAG LLM package.

dworthen:
Add GraphRAG LLM package.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -15,6 +15,18 @@
 			"console": "integratedTerminal"
 		},
 		{
+			"name": "Update",
+			"type": "debugpy",
+			"request": "launch",
+			"module": "graphrag",
+			"args": [
+				"update",
+				"--root", 
+				"${input:root_folder}"
+			],
+			"console": "integratedTerminal"
+		},
+		{
 			"name": "Query",
 			"type": "debugpy",
 			"request": "launch",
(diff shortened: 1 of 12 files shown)


packages/graphrag/graphrag/index/operations/extract_graph/graph_extractor.py
```diff
@@ -77,35 +83,40 @@ async def __call__(
...
-            }),
+            })
+        )
+
+        response = await self._model.completion_async(
+            messages=messages_builder.build(),
         )
-        results = response.output.content or ""
+        results = await gather_completion_response_async(response)
+        messages_builder.add_assistant_message(results)
```
natoverse: The builder pattern feels a bit heavyweight for our core needs. Can we make this optional for "advanced use" and expose a more straightforward basic chat method?

packages/graphrag/graphrag/index/workflows/create_community_reports.py
```diff
@@ -78,8 +81,8 @@ async def run_workflow(
         prompt=prompts.graph_prompt,
         max_input_length=config.community_reports.max_input_length,
         max_report_length=config.community_reports.max_length,
-        num_threads=model_config.concurrent_requests,
-        async_type=model_config.async_mode,
+        num_threads=config.concurrent_requests,
```
natoverse: This looks like config on the root model. It may be different per workflow. I think we discussed moving it out of llm config and into the workflow blocks?

packages/graphrag/graphrag/index/text_splitting/text_splitting.py
```diff
@@ -9,9 +9,9 @@
 from typing import cast
 
 import pandas as pd
+from graphrag_llm.tokenizer import Tokenizer
```
natoverse: I think this file was supposed to be removed with my chunking PR. Just tagging here so we have a note

PR #2189, opened 2026-01-26
add TableProvider to enable future row-by-row streaming

dayesouza:
## Description

This PR implements **Phase 1** of the storage migration plan to support streaming row-by-row operations for larger datasets with reduced memory pressure. It introduces the `TableProvider` abstraction layer that sits between GraphRAG workflows and the underlying storage mechanism, enabling future database backends and streaming operations.

All indexing workflows have been migrated from direct `Storage` calls to use `TableProvider.read_dataframe()` and `write_dataframe()` methods, while maintaining backward compatibility through wrapper functions.

### Core Infrastructure

- **New**: `TableProvider` abstract base class in `graphrag-storage/tables/`
  - `read_dataframe(table_name)` - read entire table as DataFrame
  - `write_dataframe(table_name, df)` - write entire table as DataFrame
  - `has_dataframe(table_name)` - check if table exists
  - `find_tables()` - list all available tables

- **New**: `ParquetTableProvider` implementation
  - Wraps underlying `Storage` instance (file/blob/cosmos)
  - Converts between DataFrames and Parquet format
  - Handles BytesIO streaming for in-memory operations

### Context Changes

- Updated `PipelineRunContext` to include:
  - `input_table_provider: ParquetTableProvider`
  - `output_table_provider: ParquetTableProvider` 
  - `previous_table_provider: ParquetTableProvider | None` (for update mode)

- Renamed `previous_storage` → `previous_table_provider` for consistency

- **Note**: `output_storage` remains in context for JSON files (`stats.json`, `context.json`) and GraphML snapshots

**Pattern change:**
```python
# Before
documents = await load_table_from_storage("documents", context.output_storage)
await write_table_to_storage(output, "text_units", context.output_storage)

# After
documents = await context.output_table_provider.read_dataframe("documents")
await context.output_table_provider.write_dataframe("text_units", output)
```

### DIFF
--- a/.semversioner/3.0.1.json
+++ b/.semversioner/3.0.1.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "description": "Fix missing dependency.",
+      "type": "patch"
+    }
+  ],
+  "created_at": "2026-01-28T14:34:25+00:00",
+  "version": "3.0.1"
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 # Changelog
 Note: version releases in the 0.x.y range may introduce breaking changes.
 
+## 3.0.1
+
+- patch: Fix missing dependency.
+
 ## 3.0.0
 
 - major: Monorepo restructure
--- a/packages/graphrag-cache/pyproject.toml
+++ b/packages/graphrag-cache/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "graphrag-cache"
-version = "3.0.0"
+version = "3.0.1"
 description = "GraphRAG cache package."
 authors = [
     {name = "Alonso Guevara Fernández", email = "alonsog@microsoft.com"},
@@ -31,8 +31,8 @@ classifiers = [
     "Programming Language :: Python :: 3.13",
 ]
 dependencies = [
-    "graphrag-common==3.0.0",
-    "graphrag-storage==3.0.0",
+    "graphrag-common==3.0.1",
+    "graphrag-storage==3.0.1",
 ]
 
 [project.urls]
--- a/packages/graphrag-chunking/pyproject.toml
+++ b/packages/graphrag-chunking/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "graphrag-chunking"
-version = "3.0.0"
+version = "3.0.1"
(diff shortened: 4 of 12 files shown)


.semversioner/next-release/minor-20260126224712110537.json
```diff
@@ -0,0 +1,4 @@
+{
+  "type": "minor",
```
natoverse: You can specify this as a patch since it is not API-impacting (even though it does affect the package surface area)

packages/graphrag-storage/graphrag_storage/tables/__init__.py
```diff
@@ -0,0 +1,9 @@
+# Copyright (c) 2024 Microsoft Corporation.
+# Licensed under the MIT License
+
+"""Table provider module for GraphRAG storage."""
+
+from .parquet_table_provider import ParquetTableProvider
+from .table_provider import TableProvider
+
+__all__ = ["ParquetTableProvider", "TableProvider"]
```
natoverse: I would avoid exporting `ParquetTableProvider` here, but keep `TableProvider`. The pattern we've been adopting is to export the base interface and factories, but not specific impls. This allows us to eventually move to a lazy-load factory pattern so deps aren't loaded into the py runtime unless requested specifically in a pipeline config.

packages/graphrag-storage/graphrag_storage/__init__.py
```diff
@@ -10,11 +10,14 @@
     register_storage,
 )
 from graphrag_storage.storage_type import StorageType
+from graphrag_storage.tables import ParquetTableProvider, TableProvider
 
 __all__ = [
+    "ParquetTableProvider",
```
natoverse: Same as above

packages/graphrag/graphrag/utils/storage.py
```diff
@@ -4,33 +4,25 @@
...
-from io import BytesIO
 
 import pandas as pd
-from graphrag_storage import Storage
+from graphrag_storage import ParquetTableProvider, Storage
 
 logger = logging.getLogger(__name__)
 
 
 async def load_table_from_storage(name: str, storage: Storage) -> pd.DataFrame:
```
natoverse: Given that the provider reads/writes dataframes, I think these utilities can probably be deleted

PR #2205, opened 2026-02-04
Add notebook example support for each package

gaudyb:
## Description

Add multiple example notebooks to each package for better user support. 

## Checklist


dworthen:
Hi @gaudyb, Looks like the notebooks for graphrag_llm were moved to example_notebooks. Can you update `tests/notebook/test_notebooks.py` with the new location? It appears the test is currently passing when there are no notebooks to test. Can you update the test so that an error is thrown if `notebooks_list` is empty so that we don't have false positives when the tests don't run anything.

PR #2234, opened 2026-02-18
feat: add LLM-based entity resolution workflow step

ohharsen:
## Description

This PR adds **LLM-based entity resolution** to the GraphRAG indexing pipeline. Currently, entities are deduplicated by exact title match during graph extraction. This means entities with different surface forms that refer to the same real-world thing — such as "Ahab" and "Captain Ahab", or "USA" and "United States of America" — remain as separate nodes, fragmenting the knowledge graph.

Entity resolution sends all unique entity names to the LLM in a single call, asks it to identify duplicate groups, and unifies their titles **before** title grouping and description summarization. This means:
- Duplicate entities are merged before their descriptions are summarized, producing better summaries
- Fewer summarization LLM calls (no wasted calls on duplicate entities)
- All original surface forms are preserved in an `alternative_names` column

**The feature is opt-in** — disabled by default via `entity_resolution.enabled: false`, so existing pipelines are unaffected.

## Related Issues

This addresses the general problem of entity fragmentation in knowledge graphs built by GraphRAG. Entities extracted from different text chunks often use variant names for the same concept, leading to duplicate nodes and incomplete relationship graphs.

### New files

| File | Purpose |
|------|---------|
| `prompts/index/entity_resolution.py` | Default prompt template that instructs the LLM to identify duplicate entity names from a numbered list and output groups of indices |
| `config/models/entity_resolution_config.py` | Pydantic config model with fields: `enabled`, `completion_model_id`, `model_instance_name`, `prompt` |
| `index/operations/resolve_entities/__init__.py` | Package init |
| `index/operations/resolve_entities/resolve_entities.py` | Core operation: collects unique titles, sends to LLM in a single call, parses response into rename map, applies to entity titles and relationship source/target columns, stores aliases in `alternative_names` |

### Modified files

| File | Change |
|------|--------|
| `config/defaults.py` | Added `EntityResolutionDefaults` dataclass (enabled=False) and wired into `GraphRagConfigDefaults` |
| `config/models/graph_rag_config.py` | Added `entity_resolution: EntityResolutionConfig` field |
| `config/init_content.py` | Added `entity_resolution` section to default YAML template generated by `graphrag init` |
| `data_model/schemas.py` | Added `ALTERNATIVE_NAMES` constant and included it in `ENTITIES_FINAL_COLUMNS` |
| `index/workflows/extract_graph.py` | Integrated entity resolution into the `extract_graph` function — runs after raw extraction and before title grouping/summarization |
| `index/operations/finalize_entities.py` | Ensures `alternative_names` column exists even when resolution is disabled |
| `index/update/entities.py` | Handles `alternative_names` in incremental update aggregation |

### Pipeline position

Entity resolution is integrated directly into `extract_graph`, running between raw entity extraction and title grouping/summarization:

```
extract entities from text → resolve entities (LLM) → group by title → summarize descriptions
```

This is a deliberate architectural choice — resolving before grouping means all descriptions for the same real-world entity are summarized together, producing higher quality summaries with fewer LLM calls.

### Usage

Enable in `settings.yaml`:

```yaml
entity_resolution:
  enabled: true
  # completion_model_id: default_completion_model  # optional
```

## Checklist

## Additional Notes

- The prompt uses a numbered-list + index-output format for reliable parsing. The LLM returns lines like `3, 17` meaning entities 3 and 17 are the same, with entity 3's name as canonical.
- All unique entity names are sent in a single LLM call (no batching) for simplicity and to avoid cross-batch blind spots where duplicates in different batches would never be compared.
- If the LLM call fails, the operation degrades gracefully to a no-op rather than crashing the pipeline.
- Each canonical entity gets an `alternative_names` column listing all of its original surface forms, so downstream consumers (search, UI) can still find entities by any of their names.
- The operation copies DataFrames before mutation to avoid side effects on the input data.

### DIFF
--- a/packages/graphrag/graphrag/index/operations/resolve_entities/resolve_entities.py
+++ b/packages/graphrag/graphrag/index/operations/resolve_entities/resolve_entities.py
@@ -72,8 +72,8 @@ async def resolve_entities(
     formatted_prompt = prompt.format(entity_list=entity_list)
 
     try:
-        response = await model(formatted_prompt)
-        raw = (response or "").strip()
+        response = await model.completion_async(messages=formatted_prompt)
+        raw = (response.content or "").strip()
     except Exception as e:
         logger.warning("Entity resolution LLM call failed, skipping resolution: %s", e, exc_info=True)
         return entities, relationships


ohharsen:

ohharsen:
@gaudyb could you please review this?

ohharsen:
@natoverse @dayesouza

PR #2238, opened 2026-02-20
Streaming create_final_text_units

dayesouza:
Streaming Performance Improvement (with .txt):

run time: -68.77%
peak memory: -91.99%
memory delta: -92.67%

This pull request refactors the `create_final_text_units` workflow to enable efficient streaming processing and row-level type coercion, replacing the previous DataFrame-based approach. It introduces new row transformer utilities, updates the workflow to operate on streamed table rows, and adds robust type handling for input data.

**Major workflow refactor and streaming support:**

* The `create_final_text_units` workflow is rewritten to stream rows from input tables, enrich them with related entity, relationship, and covariate IDs, and write them incrementally to the output table, instead of using pandas DataFrames for bulk processing. This enables support for larger datasets and reduces memory usage.

**Row-level type coercion utilities:**

* A new module, `row_transformers.py`, is added with a set of row-level transformer functions (e.g., `transform_entity_row`, `transform_relationship_row`, etc.) that coerce raw dictionary rows from CSV or table sources into properly typed Python dictionaries. These are now used throughout the workflow for robust data handling.

**Internal API and helper updates:**

* The workflow now uses async context managers to open tables with appropriate row transformers, and introduces helper functions (`_build_multi_ref_map`, `_build_covariate_map`) to efficiently build reverse lookup maps for enrichment.
* The `_split_list_column` function is renamed to `split_list_column` in `dfs.py`, with a backward-compatible alias to support both internal and external callers.

**Other changes:**

* A patch release note is added to document the new streaming support for `create_final_text_units`.

### DIFF
--- a/packages/graphrag/graphrag/data_model/dfs.py
+++ b/packages/graphrag/graphrag/data_model/dfs.py
@@ -36,7 +36,7 @@ def _safe_int(series: pd.Series, fill: int = -1) -> pd.Series:
 def split_list_column(value: Any) -> list[Any]:
     r"""Split a column containing a list string into an actual list.
 
-    Handles two CSV serialisation formats:
+    Handles two CSV serialization formats:
     - Comma-separated (standard ``str(list)``): ``"['a', 'b']"``
     - Newline-separated (pandas ``to_csv`` of numpy arrays):
       ``"['a'\\n 'b']"``
@@ -48,8 +48,8 @@ def split_list_column(value: Any) -> list[Any]:
     if isinstance(value, str):
         if not value:
             return []
-        normalised = value.replace("\n", ",")
-        items = [item.strip("[] '\"") for item in normalised.split(",")]
+        normalized = value.replace("\n", ",")
+        items = [item.strip("[] '\"") for item in normalized.split(",")]
         return [item for item in items if item]
     return value
 


packages/graphrag/graphrag/index/workflows/create_final_text_units.py
```diff
@@ -22,103 +28,111 @@ async def run_workflow(
...
 
-
-def _join(left, right):
-    return left.merge(
-        right,
-        on="id",
-        how="left",
-        suffixes=["_1", "_2"],
-    )
+    # Buffer all input rows before writing.  The read and write handles
```
natoverse: This is an interesting challenge, but I'm not convinced this is the right way to solve it.

The entities, relationships, and covariates are all read-only, no problem here. Text units are read/write, so that is a challenge. In a DB environment, you could imagine this being simple to handle because you aren't streaming a single file. One solution here could be that if the CSVTable is opened in read/write mode, it creates a temp copy to read from and writes (truncate) to the final file?

natoverse: Or, maybe a cleaner way is: stream the writes to a temp file, then copy it back over the original on close()

dayesouza: Done! Way cleaner :)

PR #2259, opened 2026-03-01
fix: preserve nodes with unassigned community level in filtering

### LINKED ISSUE #1808: [Fatal Bug]: Nodes Missing Community and Level Attributes Are Excluded During Query Execution

### Describe the bug

Image
In **V1.0.0**, after running the Leiden algorithm in `layout_graph`, some nodes were not assigned a **community** and **level**. To handle this, the code assigned **pseudo community and level** as a fallback mechanism. While this approach was not ideal, it did not introduce a **fatal bug**, as the missing nodes were still retained.

Image
In **V2.0.0**, the handling of nodes changed:  
- Nodes are no longer **saved persistently** but are instead **generated dynamically** during queries.  
- The previous fallback mechanism (assigning pseudo community and level) was **removed**, meaning that nodes without a level attribute (`level=None`) are left unprocessed.  
- This causes an issue in `_filter_under_community_level`: nodes with `level=None` are **wrongly discarded**, leading to an approximately **10% data loss** in the graph.  

To confirm the issue, I saved `nodes_df` before and after `_filter_under_community_level` and observed that these unassigned nodes were filtered out incorrectly.

This results in significant data loss and affects query results, making it a **critical bug** that needs to be addressed.

## Workaround

To mitigate the issue temporarily, you can add the following code snippet before the filtering step. This workaround assigns default values to missing `level` and `community` attributes, ensuring that nodes are not inadvertently discarded:

```python
nodes_df["level"] = nodes_df["level"].fillna(0)
nodes_df["level"] = nodes_df["level"].astype(int)
nodes_df["community"] = nodes_df["community"].fillna(-1)
nodes_df["community"] = nodes_df["community"].astype(int)
```
Screenshot

### Steps to reproduce

1. Create or load a graph that includes isolated nodes (nodes not connected to any other nodes).
2. Run Index
3. Execute local search
4. Observe that nodes with a missing level attribute (level=None) are discarded during the filtering process, resulting in approximately 10% of nodes being lost.

Image

### Expected Behavior

- All nodes, including isolated ones, should either be assigned a valid community and level or be handled in a way that retains them in the dataset.
- The filtering function should not discard nodes with a missing level attribute unless explicitly intended, thereby preventing unintended data loss.

### GraphRAG Config Used

_No response_

### Logs and screenshots

_No response_

### Additional Information

- GraphRAG Version: V2.0.0
- Python Version: 3.12.9
- Related Issues: None, but I also issued another fatal bug in #1718

Br1an67:
## Description

Fix `_filter_under_community_level` to retain nodes where `level` is NaN, which represent isolated nodes not assigned to any community by the Leiden algorithm. Previously these nodes were silently dropped because NaN comparisons always evaluate to False in pandas, causing approximately 10% data loss.

## Related Issues

Fixes #1808

## Proposed Changes

- Add `| df.level.isna()` condition to `_filter_under_community_level` so NaN-level nodes are preserved
- Add unit tests verifying both NaN preservation and correct level filtering

## Checklist

## Additional Notes

Downstream code already handles NaN community values with `fillna(-1)`, so preserving these nodes is consistent with the existing data flow.

### DIFF
--- a/packages/graphrag/graphrag/query/indexer_adapters.py
+++ b/packages/graphrag/graphrag/query/indexer_adapters.py
@@ -221,5 +221,5 @@ def _filter_under_community_level(
 ) -> pd.DataFrame:
     return cast(
         "pd.DataFrame",
-        df[df.level <= community_level],
+        df[(df.level <= community_level) | df.level.isna()],
     )
--- a/tests/unit/query/test_filter_community_level.py
+++ b/tests/unit/query/test_filter_community_level.py
@@ -1,0 +1,42 @@
+# Copyright (c) 2024 Microsoft Corporation.
+# Licensed under the MIT License
+
+"""Tests for _filter_under_community_level preserving NaN-level nodes."""
+
+import numpy as np
+import pandas as pd
+
+from graphrag.query.indexer_adapters import _filter_under_community_level
+
+
+def test_filter_preserves_nan_level_nodes():
+    """Nodes with level=NaN should not be discarded by the filter.
+
+    Regression test for issue #1808 where isolated nodes without a community
+    assignment (level=None) were incorrectly dropped.
+    """
+    df = pd.DataFrame({
+        "id": ["a", "b", "c", "d"],
+        "level": [0, 1, 2, np.nan],
+        "community": [1, 2, 3, np.nan],
+    })
+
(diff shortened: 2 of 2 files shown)


Br1an67:
Hi — just a gentle bump on this. Happy to make any changes if needed!

PR #2278, opened 2026-03-12
Fix tool-call fallback for structured completion content

### LINKED ISSUE #2263: [Bug]: `model_supports_json` is no longer supported on v3, breaking custom model setups

- [ ] My model is hosted on OpenAI or Azure. If not, please look at the "model providers" issue and don't file a new one here.

### Describe the bug

Most models do not support structured output. In this case the LLM caller (liteLLM) usually falls back to a tool call. However, a tool call will cause response content to be empty, which then the LLM caller tries to parse, and fail.

In v2 there is a `model_supports_json` param, which can be disabled to not use structured output in the LLM calls, to avoid this issue. In v3 this setting is gone, and so every model that doesn't support structured output will fail to work.

### Steps to reproduce

_No response_

### Expected Behavior

_No response_

### GraphRAG Config Used

```yaml
# Paste your config here

```

### Logs and screenshots

_No response_

### Additional Information

- GraphRAG Version: 3.0.5

suger-m:
## Summary
- Fix structured response parsing when provider responses return empty assistant content and place JSON in function tool-call arguments.
- Add fallback in LLMCompletionResponse.content to read tool_calls[].function.arguments when message content is empty.
- Update gather_completion_response and gather_completion_response_async to use response.content.
- Add unit regression tests for content fallback, gather helpers, and structured parsing.

## Issue
Fixes #2263

## Validation
- uvx ruff check packages/graphrag-llm/graphrag_llm/types/types.py packages/graphrag-llm/graphrag_llm/utils/gather_completion_response.py tests/unit/utils/test_completion_response.py
- python -m pytest tests/unit/utils/test_completion_response.py -q

### DIFF
--- a/.semversioner/next-release/patch-20260312083316079043.json
+++ b/.semversioner/next-release/patch-20260312083316079043.json
@@ -1,0 +1,4 @@
+{
+  "type": "patch",
+  "description": "Fix structured response parsing when providers return tool calls with empty message content"
+}


suger-m:

suger-m:
@microsoft-github-policy-service agree company="your company"

suger-m:
@AlonsoGuevara

PR #2354, opened 2026-05-11
feat: native CosmosTableProvider with namespace partitioning

disamhembere:
## Summary

Replaces the parquet-decomposition approach in `AzureCosmosStorage` with a native `CosmosTableProvider` that implements `TableProvider` directly against Cosmos DB.

## Problem

The current Cosmos DB integration forces all data through a parquet serialization layer designed for file/blob storage, causing:

- **4 serde hops** per read/write (DataFrame → parquet → DataFrame → JSON → Cosmos, reversed on read)
- **Cross-partition fan-out** on every multi-row query (partition key is `/id`, so each document is its own partition)
- **Entity-specific hacks** in the storage layer (`if prefix == "entities":`)
- **Broken `child()`** — no-op, so update-run delta/previous isolation doesn't work
- **`clear()` drops the entire database**
- **Sync SDK** used inside async methods (blocks event loop)
- **Unparameterized f-string queries**

## Solution

| | Before | After |
|---|---|---|
| Serde hops | 4 per read/write | 1 |
| Query pattern | Cross-partition fan-out | Single-partition |
| Entity hacks in storage | Yes | None |
| Update-run isolation | Broken | Namespace partitioning |
| SDK | Sync (blocks event loop) | Async |
| Query safety | f-string interpolation | Parameterized |
| Write strategy | Row-by-row upsert | Transactional batch (default 50, max 100) |

### Architecture

- **`CosmosTableProvider`** implements `TableProvider` directly — DataFrame rows become Cosmos documents natively. Partition key is `/namespace`, co-locating all data for a logical scope in a single partition.
- **`CosmosTable`** implements `Table` for streaming row access with async SDK and server-side pagination.
- **`AzureCosmosStorage`** simplified to key-value only (context.json, stats.json, cache). Working `child()` via `:` namespace prefix.
- **`TableProvider.child()`** — new non-abstract method for namespace isolation. `ParquetTableProvider`/`CSVTableProvider` delegate to `Storage.child()`.
- **Transactional batch writes** — configurable `batch_size` (default 50, max 100) with automatic fallback to individual upserts on batch failure.

### Migration

- **Legacy fallback reads**: when `legacy_container` is configured, reads transparently from old-format containers while writing exclusively to the new schema.
- **Re-index migrates data** automatically (reads from legacy, writes to new).
- **CLI tool** (`graphrag migrate-cosmos`) planned for migration without re-indexing.

## Changes

| File | Change |
|------|--------|
| `graphrag_storage/tables/table_provider.py` | Add `child()` default method |
| `graphrag_storage/tables/table_type.py` | Add `CosmosDB` enum |
| `graphrag_storage/tables/table_provider_config.py` | Add Cosmos + batch_size fields |
| `graphrag_storage/tables/cosmos_table_provider.py` | **New** — native Cosmos table provider |
| `graphrag_storage/tables/cosmos_table.py` | **New** — streaming Table impl |
| `graphrag_storage/tables/table_provider_factory.py` | Wire `cosmosdb` case |
| `graphrag_storage/tables/parquet_table_provider.py` | Add `child()` |
| `graphrag_storage/tables/csv_table_provider.py` | Add `child()` |
| `graphrag_storage/azure_cosmos_storage.py` | Simplified to key-value only |
| `graphrag/index/run/run_pipeline.py` | Use `table_provider.child()` |
| `graphrag/index/run/utils.py` | Refactor `get_update_table_providers` |

## Testing

- 302 unit tests + 15 verb tests pass (no regressions)
- pyright: 0 errors, ruff: all checks pass
- E2E tested against Cosmos DB Linux emulator (vNext, ARM64):
  - CosmosTableProvider lifecycle (write, read, has, list, child, open, stream, truncate, batch)
  - AzureCosmosStorage key-value (set, get, has, child, keys, delete, clear)
  - Factory wiring, update-run simulation, full migration scenario

## Config

```yaml
table_provider:
  type: cosmosdb
  account_url: https://myaccount.documents.azure.com:443/
  database_name: graphrag
  container_name: graphrag-tables
  batch_size: 50              # default; max 100
  legacy_container: graphrag-output  # optional: enables migration fallback
```

image

### DIFF
--- a/packages/graphrag-storage/graphrag_storage/tables/cosmos_table_provider.py
+++ b/packages/graphrag-storage/graphrag_storage/tables/cosmos_table_provider.py
@@ -1,0 +1,453 @@
+# Copyright (c) 2024 Microsoft Corporation.
+# Licensed under the MIT License
+
+"""Cosmos DB native table provider implementation.
+
+Stores each DataFrame row as a Cosmos document, eliminating the
+parquet serialisation round-trip used by ParquetTableProvider.
+
+Container schema
+----------------
+Partition key : /namespace
+Document id   : {table_name}:{row_key}
+
+Every document carries two internal fields used for routing:
+    namespace  – partition key, set by child() hierarchy
+    table_name – discriminator for per-table queries
+
+All queries target a single namespace partition (no cross-partition fan-out).
+"""
+
+from __future__ import annotations
+
+import json
+import logging
+from typing import Any
+
+import pandas as pd
+from azure.cosmos.aio import ContainerProxy, CosmosClient, DatabaseProxy
+from azure.cosmos.partition_key import PartitionKey
+from azure.identity.aio import DefaultAzureCredential
+
+from graphrag_storage.tables.cosmos_table import (
+    CosmosTable,
+    _strip_cosmos_metadata,
+)
+from graphrag_storage.tables.table import RowTransformer, Table
+from graphrag_storage.tables.table_provider import TableProvider
+
+logger = logging.getLogger(__name__)
+
+_DEFAULT_PAGE_SIZE = 100
+
+# Cosmos system / internal fields to exclude from DataFrame columns.
+_INTERNAL_FIELDS = frozenset({
+    "_rid",
+    "_self",
+    "_etag",
+    "_attachments",
+    "_ts",
+    "namespace",
+    "table_name",
+})
+
+
+class CosmosTableProvider(TableProvider):
+    """TableProvider backed by Azure Cosmos DB (NoSQL API, async SDK).
+
+    Each table is stored as a set of documents within a single container,
+    discriminated by the ``table_name`` field.  Namespace isolation (used
+    by the update pipeline's delta / previous split) is achieved via the
+    ``/namespace`` partition key.
+    """
+
+    # ------------------------------------------------------------------
+    # Construction
+    # ------------------------------------------------------------------
+
+    def __init__(
+        self,
+        *,
+        connection_string: str | None = None,
+        account_url: str | None = None,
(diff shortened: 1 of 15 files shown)


packages/graphrag-storage/graphrag_storage/tables/csv_table_provider.py
```diff
@@ -138,3 +138,13 @@ def open(
             truncate=truncate,
             encoding=encoding,
         )
+
+    def child(self, name: str | None) -> "CSVTableProvider":
```
natoverse: Does adding the `child` methods solve a common pain point versus creating the child storage separately?

packages/graphrag-storage/graphrag_storage/tables/cosmos_table_provider.py
```diff
@@ -0,0 +1,479 @@
...
+        """Close the underlying Cosmos client if we own it."""
+        if self._owns_client and self._cosmos_client is not None:
+            await self._cosmos_client.close()
+            self._cosmos_client = None
+
+    # ------------------------------------------------------------------
+    # Internal helpers
+    # ------------------------------------------------------------------
+
+    async def _delete_table(self, container: ContainerProxy, table_name: str) -> None:
```
dayesouza, line 316: In here would it be worth batching, like in writes?

dayesouza (approved):

packages/graphrag-storage/graphrag_storage/tables/cosmos_table_provider.py
disamhembere: Good feedback, done!

PR #2363, opened 2026-05-23
fix(config): preserve literal $ in config values

### LINKED ISSUE #2349: Bug: `string.Template.substitute()` crashes on `$` in config values like regex `file_pattern`
## Description

The config loader in `graphrag-common` uses `string.Template.substitute()` to process environment variables in `settings.yaml`. However, `substitute()` treats **every** `$` character as a placeholder prefix — including `$` that happens to be part of a regex pattern or any other string value. This causes a hard crash with an unhelpful `ValueError`.

## Reproduction

Create a `settings.yaml` with:

```yaml
input:
  type: markitdown
  file_pattern: ".*\\.md$"
```

Run `graphrag index --root <dir>`.

## Error

```
ValueError: Invalid placeholder in string: line 50, col 25
```

This happens because `$` at the end of `.*\.md$` (a valid regex anchor) is treated by `string.Template` as a template placeholder, and a bare `$` without a valid identifier causes `substitute()` to raise `ValueError`.

## Root Cause

**File:** `graphrag-common/graphrag_common/config/load_config.py`, the `_parse_env_variables` function:

```python
def _parse_env_variables(text: str) -> str:
    """Parse environment variables in the configuration text."""
    try:
        return Template(text).substitute(os.environ)
    except KeyError as error:
        msg = f"Environment variable not found: {error}"
        raise ConfigParsingError(msg) from error
```

`substitute()` raises `ValueError` for any invalid placeholder (not just `KeyError` for missing env vars). So bare `$` in regex patterns, `$VAR` with special characters, etc. all crash the config loader.

This affects:
- `file_pattern` with regex ending in `$` (very common, e.g. `.*\.md$`)
- Any other config value that happens to contain `$`
- `prompt` paths containing `$` characters

## Proposed Fix

Replace `substitute()` with `safe_substitute()`, which silently leaves unrecognized placeholders as literal text instead of raising an error:

```python
def _parse_env_variables(text: str) -> str:
    """Parse environment variables in the configuration text."""
    return Template(text).safe_substitute(os.environ)
```

`safe_substitute()` handles `$$` escape sequences but does **not** raise an error on bare `$` or missing keys — it leaves them unchanged as literal text. This is the expected behavior: a `$` in a regex pattern should stay as `$`, not crash the config loader.

A `KeyError` or `ValueError` from `substitute()` is the wrong place to validate config values anyway — Pydantic schema validation in the config models already handles that properly.

## Environment

- graphrag version: 3.0.9
- Python: 3.12

jluocsa:
## Description

Fixes #2349.

`_parse_env_variables` in `graphrag-common` calls
`string.Template(text).substitute(os.environ)`. `substitute()` treats
**every** `$` character as a placeholder prefix and raises
`ValueError: Invalid placeholder in string` for any `$` that is not
followed by a valid Python identifier or `{ident}` / `$$` escape. Common
config values fall into that trap, most notably regex anchors like
`file_pattern: ".*\\.md$"`.

### Reproduction

```yaml
# settings.yaml
input:
  type: markitdown
  file_pattern: ".*\\.md$"
```

```
$ graphrag index --root <dir>
ValueError: Invalid placeholder in string: line 50, col 25
```

### Fix

Swap `substitute()` for `safe_substitute()`. `safe_substitute()` leaves
unrecognized `$` sequences as literal text, which is the expected
behaviour for arbitrary string config values.

To keep the existing "missing env var raises `ConfigParsingError`"
contract (and the test that exercises it), the function now also
inspects `Template.get_identifiers()` (Python 3.11+, which is graphrag's
minimum) and raises `ConfigParsingError` up-front if any referenced
identifier has no corresponding entry in `os.environ`.

Net result:

| Input | Before | After |
| --- | --- | --- |
| `file_pattern: ".*\\.md$"` | `ValueError` crash | preserved literally |
| `val: $$literal` | `$literal` | `$literal` (unchanged) |
| `val: $MY_VAR` (set) | substituted | substituted |
| `val: ${MISSING}` | `ConfigParsingError` | `ConfigParsingError` (unchanged) |

### Tests

- New unit test `test_load_config_preserves_literal_dollar_sign`
  covering bare `$` in a regex-like value, `$$` escape, and a mixed
  payload that combines a regex `$` with a real env-var reference.
- Existing `test_load_config_validation` (which expects
  `ConfigParsingError` for a missing env var) still passes — the
  `get_identifiers()` pre-check keeps that path intact.

Local run:

```
.venv\Scripts\python.exe -m pytest tests/unit/load_config -q
... 3 passed in 1.98s
```

`ruff check` and `ruff format --check` are clean on the changed files.

### Notes

- Changelog: added `.semversioner/next-release/patch-…json` per
  `CONTRIBUTING.md`.
- Diff: +1 source file, +1 test, +1 semversioner entry. No behavioural
  change for any code path that was previously succeeding.


dworthen:
`$` need to be escaped with another `$`, e.g., `.*\\.md$$`

PR #2372, opened 2026-05-30
fix: add missing return type hints on generator functions in text_utils.py (fixes #2330)

### LINKED ISSUE #2330: Missing return type hints on generator functions in query/llm/text_utils.py
## Summary

The two generator functions in `packages/graphrag/graphrag/query/llm/text_utils.py` are missing return type annotations:

- `batched(iterable: Iterator, n: int)` — yields `tuple`, return type not declared
- `chunk_text(text: str, max_tokens: int, tokenizer: Tokenizer | None = None)` — yields `str`, return type not declared

`Iterator` is already imported from `collections.abc` in this file, so no new imports are needed.

## Proposed Fix

Add `-> Iterator[tuple]` and `-> Iterator[str]` to the respective signatures. This improves downstream type safety for static analysis tools (`mypy`, `pyright`) and is consistent with the annotation style of `try_parse_json_object` in the same file.

## Notes

- Annotation-only change, zero runtime impact
- Verified locally with `mypy --ignore-missing-imports`

hanhan761:
## Summary

Add missing return type hints to two generator functions in \	ext_utils.py\.

## Changes

- \atched(iterable, n)\: \-> Iterator[tuple]\
- \chunk_text(text, max_tokens, tokenizer)\: \-> Iterator[str]\

## Issue

Fixes #2330

## Verification

- Annotation-only change, zero runtime impact
- \collections.abc.Iterator\ already imported

### DIFF
--- a/packages/graphrag/graphrag/query/llm/text_utils.py
+++ b/packages/graphrag/graphrag/query/llm/text_utils.py
@@ -17,7 +17,7 @@ from graphrag.tokenizer.get_tokenizer import get_tokenizer
 logger = logging.getLogger(__name__)
 
 
-def batched(iterable: Iterator, n: int):
+def batched(iterable: Iterator, n: int) -> Iterator[tuple]:
     """
     Batch data into tuples of length n. The last batch may be shorter.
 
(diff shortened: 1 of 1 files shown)


gaudyb:
Fixed in #2443

Issue #2265, opened 2026-03-04
[Bug]: Pipeline error: Column 1 named vector expected length 44 but got length 11


JieHan17:
### Do you need to file an issue?

- [ ] I have searched the existing issues and this bug is not already filed.
- [ ] My model is hosted on OpenAI or Azure. If not, please look at the "model providers" issue and don't file a new one here.
- [ ] I believe this is a legitimate bug, not just a question. If this is a question, please use the Discussions area.

### Describe the bug

Starting workflow: generate_text_embeddings
[2026-03-04T04:22:36Z WARN  lance::dataset::write::insert] No existing dataset at /home/ms_graphrag/secreq_ai_sample/output/lancedb/entity_description.lance, it will be created
Pipeline error: Column 1 named vector expected length 44 but got length 11..........................
Pipeline complete

### Steps to reproduce

The embedded model uses nomic-embed-text-v2-moe deployed locally via lmstudio.

### GraphRAG Config Used

```yaml
# Paste your config here
embedding_models:
  default_embedding_model:
    model_provider: openai
    auth_method: api_key
    api_key: "lmstudio" #${GRAPHRAG_API_KEY}
    model: nomic-embed-text-v2-moe
    api_base: http://localhost:1234/v1
    retry:
      type: exponential_backoff

vector_store:
  type: lancedb
  db_uri: output/lancedb
  community_full_content:
    vector_size: 768
  entity_description:
    vector_size: 768
  text_unit_text:
    vector_size: 768
```

### Logs and screenshots

[7 rows x 15 columns]
2026-03-04 12:09:55.0252 - INFO - graphrag.index.workflows.generate_text_embeddings - Workflow started: generate_text_embeddings
2026-03-04 12:09:55.0278 - INFO - graphrag.index.workflows.generate_text_embeddings - Embedding the following fields: ['entity_description', 'community_full_content', 'text_unit_text']
2026-03-04 12:09:55.0692 - INFO - graphrag.index.operations.embed_text.run_embed_text - embedding 44 inputs via 44 snippets using 3 batches. max_batch_size=16, batch_max_tokens=8191
2026-03-04 12:09:55.0796 - INFO - graphrag.logger.progress - generate embeddings progress: 1/3
2026-03-04 12:09:55.0842 - INFO - graphrag.logger.progress - generate embeddings progress: 2/3
2026-03-04 12:09:55.0878 - INFO - graphrag.logger.progress - generate embeddings progress: 3/3
[...]

PR #2378, opened 2026-05-30
fix: validate embedding vector dimensions before vector store insertion (ref #2265)

hanhan761:
Validate embedding vector dimensions before passing data to the vector store, so inconsistent dimensions produce a clear error message instead of a cryptic LanceDB error.

Related to #2265 - users hit this when their embedding model returns vectors of different lengths, which makes it impossible to insert into a fixed-dimension index.


gaudyb:
Fixed in PR #2442

PR #2388, opened 2026-05-30
fix: add missing return type hints on generator functions

hanhan761:
## Summary\n\nAdd \-> Iterator[tuple]\ and \-> Iterator[str]\ return type annotations to \atched()\ and \chunk_text()\ respectively.\n\n## Issue\n\nFixes #2330\n\n## Verification\n\n- Annotation-only change, zero runtime impact\n- \Iterator\ already imported


gaudyb:
Fixed in #2443

PR #2390, opened 2026-06-01
fix(graphrag-llm): accept non-OpenAI service_tier values in LLMCompletionResponse

### LINKED ISSUE #2389: LLMCompletionResponse rejects non-OpenAI service_tier values from LiteLLM providers
### Describe the bug

When using GraphRAG with non-OpenAI LLM providers routed through LiteLLM (e.g. Gemini via OpenRouter), the `extract_graph` workflow fails with a pydantic validation error:

```
Error: 1 validation error for LLMCompletionResponse
service_tier
  Input should be 'auto', 'default', 'flex', 'scale' or 'priority'
  For further information visit https://errors.pydantic.dev/2.13/v/literal_error
```

### Root cause

`LLMCompletionResponse` in `graphrag_llm/types/types.py` extends `openai.types.chat.ChatCompletion`, which defines:

```python
service_tier: Optional[Literal["auto", "default", "flex", "scale", "priority"]] = None
```

In `graphrag_llm/completion/lite_llm_completion.py`, the LiteLLM `ModelResponse` is converted via:

```python
return LLMCompletionResponse(**response.model_dump())
```

When the upstream provider returns a `service_tier` value outside the OpenAI literal (e.g. Gemini via OpenRouter returns a non-standard string), pydantic rejects it during model construction.

### Steps to reproduce

1. Configure GraphRAG to use a non-OpenAI model via LiteLLM (e.g. `gemini/gemini-2.5-flash-preview` through OpenRouter)
2. Run indexing on any document
3. The `create_base_text_units` workflow succeeds, but `extract_graph` fails with the validation error above

### Expected behavior

`service_tier` is an OpenAI-specific field that GraphRAG does not use. Non-standard values should not cause indexing to fail.

### Suggested fix

Either:

1. **Widen the type** — override `service_tier` in `LLMCompletionResponse` to accept `str | None` instead of the strict literal
2. **Strip the field** — pop `service_tier` from `response.model_dump()` before constructing `LLMCompletionResponse` in `lite_llm_completion.py`
3. **Use `model_validate` with strict=False** or configure the model to ignore extra/non-conforming fields

Option 2 is the most minimal:

```python
# In _base_completion and _base_completion_async:
if isinstance(response, ModelResponse):
    dump = response.model_dump()
    dump.pop("service_tier", None)
    return LLMCompletionResponse(**dump)
```

### Environment

- graphrag >= 2.0
- graphrag-llm (latest as of May 2026)
- LiteLLM routing to Gemini via OpenRouter
- Python 3.12
- pydantic 2.13

### Workaround

Monkey-patch `LLMCompletionResponse` at application startup:

```python
from graphrag_llm.types import LLMCompletionResponse
LLMCompletionResponse.model_fields["service_tier"].annotation = str | None
LLMCompletionResponse.model_rebuild()
```

LijuanTang94:
Closes #2389

## Problem

Non-OpenAI providers routed through LiteLLM (e.g. Gemini via OpenRouter) return a `service_tier` value outside OpenAI's strict literal (`auto`/`default`/`flex`/`scale`/`priority`). Because `LLMCompletionResponse` extends `openai.types.chat.ChatCompletion`, constructing it via `LLMCompletionResponse(**response.model_dump())` in `lite_llm_completion.py` fails with a pydantic `literal_error`, breaking the `extract_graph` workflow:

```
1 validation error for LLMCompletionResponse
service_tier
  Input should be 'auto', 'default', 'flex', 'scale' or 'priority'
```

## Fix

`service_tier` is an OpenAI-specific field that graphrag-llm does not use. This overrides it in the `LLMCompletionResponse` subclass to accept `str | None`, so non-standard provider values no longer fail validation. This is more robust than stripping the field on a single conversion path, as it covers every place the model is constructed.

## Verification

- Constructing `LLMCompletionResponse` with a non-OpenAI `service_tier` (e.g. an OpenRouter value) now succeeds instead of raising.
- Standard OpenAI values (`default`) and `None` still work.
- The `content` computed field is unaffected.
- `ruff check` / `ruff format --check` pass; semversioner change file added.

### DIFF
--- a/.semversioner/next-release/patch-2026053119245844167500.json
+++ b/.semversioner/next-release/patch-2026053119245844167500.json
@@ -1,0 +1,4 @@
+{
+  "type": "patch",
+  "description": "Allow non-OpenAI service_tier values in LLMCompletionResponse so indexing does not fail with LiteLLM-routed providers (e.g. Gemini via OpenRouter)."
+}
--- a/packages/graphrag-llm/graphrag_llm/types/types.py
+++ b/packages/graphrag-llm/graphrag_llm/types/types.py
@@ -91,6 +91,12 @@ class LLMCompletionResponse(ChatCompletion, Generic[ResponseFormat]):
     provided ResponseFormat model.
     """
 
+    # Widen OpenAI's strict `service_tier` literal. Non-OpenAI providers routed
+    # through LiteLLM (e.g. Gemini via OpenRouter) return values outside OpenAI's
+    # enum, which would otherwise fail pydantic validation when constructing this
+    # model. graphrag-llm does not use this field. See issue #2389.
+    service_tier: str | None = None  # type: ignore
+
     formatted_response: ResponseFormat | None = None  # type: ignore
     """Formatted response according to the specified response_format json schema."""
 


LijuanTang94:

LijuanTang94:
Hi team — just checking in on this bug fix. Happy to update if anything is needed!

PR #2391, opened 2026-06-01
Fix whitespace-only LanceDB db_uri fallback

### LINKED ISSUE #2381: bug: vector store db_uri whitespace check uses .strip without parentheses, always False
## Description

In `graph_rag_config.py`, the vector store config validation checks `store.db_uri.strip == ""` to detect empty URIs. However, `.strip` is used without parentheses, so it returns a bound method reference instead of the trimmed string. The comparison `bound_method == ""` is always `False`, so whitespace-only URIs are never caught and the default value is never applied.

## Root Cause

Line 273: `if not store.db_uri or store.db_uri.strip == "":`

`store.db_uri.strip` is the method object `str.strip`, not the result of `strip()`. This method object is never equal to `""`, so the guard never triggers for whitespace-only values.

## Proposed Fix

Change `store.db_uri.strip == ""` to `store.db_uri.strip() == ""`.

yuzhiyang1:
## Description

Fixes LanceDB vector store db_uri validation so whitespace-only values fall back to the default db_uri instead of resolving to the current working directory.

## Related Issues

Fixes #2381

## Proposed Changes

- Call `strip()` when checking whether the LanceDB db_uri is blank.
- Add a unit test covering whitespace-only db_uri values.
- Add a semversioner patch change entry.

## Checklist

## Additional Notes

Validation run locally:

- `uv run pytest tests/unit/config/test_config.py -q`
- `uv run ruff format packages/graphrag/graphrag/config/models/graph_rag_config.py tests/unit/config/test_config.py --check`
- `uv run ruff check packages/graphrag/graphrag/config/models/graph_rag_config.py tests/unit/config/test_config.py`

### DIFF
--- a/.semversioner/next-release/patch-20260601101913537242.json
+++ b/.semversioner/next-release/patch-20260601101913537242.json
@@ -1,0 +1,4 @@
+{
+  "type": "patch",
+  "description": "Fix whitespace-only LanceDB vector store db_uri fallback."
+}
--- a/packages/graphrag/graphrag/config/models/graph_rag_config.py
+++ b/packages/graphrag/graphrag/config/models/graph_rag_config.py
@@ -270,7 +270,8 @@ class GraphRagConfig(BaseModel):
         """Validate the vector store configuration."""
         store = self.vector_store
         if store.type == VectorStoreType.LanceDB:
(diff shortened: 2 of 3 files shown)


yuzhiyang1:

PR #2412, opened 2026-06-23
fix(config): call .strip() correctly in vector store db_uri validation

### LINKED ISSUE #2381: bug: vector store db_uri whitespace check uses .strip without parentheses, always False

himanshupandeytech-ai:
## Summary
Fixes #2381
The condition in `_validate_vector_store_db_uri()` used `store.db_uri.strip` without parentheses.
A bound method reference is always truthy, so the comparison `store.db_uri.strip == ""` was always `False`.
A whitespace-only `db_uri` was never replaced with the default and was passed directly to `Path.resolve()`.
## Fix
```python
# Before
if not store.db_uri or store.db_uri.strip == "":
# After
if not store.db_uri or store.db_uri.strip() == "":
```
## Tests
Added `tests/unit/config/test_vector_store_db_uri_validation.py` covering:
- empty string, whitespace-only, tab-only, newline-only, mixed whitespace (all fall back to default)
- None / omitted db_uri (falls back to default)
- valid relative path (resolved to absolute)
- valid absolute path (preserved)

### DIFF
--- a/.semversioner/next-release/patch-20260623010541725400.json
+++ b/.semversioner/next-release/patch-20260623010541725400.json
@@ -1,0 +1,4 @@
+{
+  "type": "patch",
+  "description": "Fix vector store db_uri whitespace validation \u2014 .strip was referenced as a method object instead of being called, causing whitespace-only URIs to bypass the default fallback"
+}
--- a/packages/graphrag/graphrag/config/models/graph_rag_config.py
+++ b/packages/graphrag/graphrag/config/models/graph_rag_config.py
@@ -270,7 +270,7 @@ class GraphRagConfig(BaseModel):
         """Validate the vector store configuration."""
         store = self.vector_store
         if store.type == VectorStoreType.LanceDB:
-            if not store.db_uri or store.db_uri.strip == "":
+            if not store.db_uri or store.db_uri.strip() == "":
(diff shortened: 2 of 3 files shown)


himanshupandeytech-ai:

manjitsithuraj (approved):

dogukanincee (approved):
LGTM

himanshupandeytech-ai:
Hey [dworthen](https://github.com/dworthen) Thanks for the review. Could you let me know why this was closed without merging? 
I want to understand if the fix was addressed elsewhere or if there is something I should have done differently. 
Happy to iterate if there is a preferred approach.

PR #2413, opened 2026-06-23
Handle missing root communities in dynamic selection

### LINKED ISSUE #2382: DynamicCommunitySelection crashes with KeyError when no communities at level 0
## Description

`DynamicCommunitySelection.__init__` at `dynamic_community_selection.py:71` accesses `self.levels["0"]` directly without checking whether the key exists. If the input data contains no communities at level "0" (e.g., all communities start at level "1" or the communities list is empty), this raises `KeyError: '0'` and crashes the dynamic community selection flow.

## Root Cause

```python
# line 71
self.starting_communities = self.levels["0"]
```

`self.levels` is built from the communities list at lines 62-68. If no community has `level="0"`, the key "0" is never added to the dict.

## Expected Behavior

When no communities exist at level 0, `starting_communities` should default to an empty list (`[]`) so the selection gracefully yields no results instead of crashing.

## Proposed Fix

Replace `self.levels["0"]` with `self.levels.get("0", [])` on line 71.

## Impact

This affects users of the dynamic community selection feature (used in DRIFT search) when their indexed data has no root-level communities, which can happen with small or domain-specific datasets.

yinli-systems:
## Summary

Fixes #2382 by allowing `DynamicCommunitySelection` to initialize when the community hierarchy has no level `0` entries. In that case there are no root communities to seed the dynamic traversal, so the selector now starts with an empty queue instead of raising `KeyError` during construction.

## Tests

- `git diff --check`
- `PYTHONPYCACHEPREFIX=/private/tmp/graphrag-pycache /usr/bin/python3 -m py_compile packages/graphrag/graphrag/query/context_builder/dynamic_community_selection.py tests/unit/query/context_builder/dynamic_community_selection.py`
- `UV_CACHE_DIR=/private/tmp/uv-cache-graphrag uv run ruff check packages/graphrag/graphrag/query/context_builder/dynamic_community_selection.py tests/unit/query/context_builder/dynamic_community_selection.py`
- `UV_CACHE_DIR=/private/tmp/uv-cache-graphrag uv run ruff format --check packages/graphrag/graphrag/query/context_builder/dynamic_community_selection.py tests/unit/query/context_builder/dynamic_community_selection.py`
- `UV_CACHE_DIR=/private/tmp/uv-cache-graphrag PYTHONPATH=packages/graphrag uv run --package graphrag pytest -q tests/unit/query/context_builder/dynamic_community_selection.py`

### DIFF
--- a/packages/graphrag/graphrag/query/context_builder/dynamic_community_selection.py
+++ b/packages/graphrag/graphrag/query/context_builder/dynamic_community_selection.py
@@ -67,8 +67,8 @@ class DynamicCommunitySelection:
             if community.short_id in self.reports:
                 self.levels[community.level].append(community.short_id)
 
-        # start from root communities (level 0)
-        self.starting_communities = self.levels["0"]
+        # start from root communities (level 0), if any
+        self.starting_communities = self.levels.get("0", [])
 
     async def select(self, query: str) -> tuple[list[CommunityReport], dict[str, Any]]:
         """
--- a/tests/unit/query/context_builder/dynamic_community_selection.py
+++ b/tests/unit/query/context_builder/dynamic_community_selection.py
@@ -5,6 +5,7 @@
 
 from unittest.mock import MagicMock
 
+import pytest
 from graphrag.data_model.community import Community
 from graphrag.data_model.community_report import CommunityReport
 from graphrag.query.context_builder.dynamic_community_selection import (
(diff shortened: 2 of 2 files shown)


yinli-systems:
Closing after re-auditing the maintainer guidance on #2382 and the earlier #2384. Treating a missing level-0 community as an empty traversal would mask an indexing invariant violation, which @natoverse explicitly identified as the more serious underlying problem. This PR therefore should not merge in its current direction.

PR #2423, opened 2026-07-05
fix(input): ignore blank lines when reading JSONL files

VectorPeak:
## Description

Fix JSONL input loading so blank or whitespace-only lines do not cause the entire `.jsonl` file to be skipped.

`JSONLinesFileReader.read_file()` currently parses every entry returned by `text.splitlines()` with `json.loads(line)`. When a `.jsonl` file contains a blank line between otherwise valid records, `json.loads("")` raises `JSONDecodeError`. That exception bubbles up to `InputReader._iterate_files()`, which logs the error and skips the whole file.

This change filters out blank or whitespace-only lines before parsing JSONL records, while preserving the existing behavior for non-empty malformed JSON lines.

## Related Issues

Fixes #2424.

## Proposed Changes

- Ignore blank or whitespace-only lines when reading JSONL input files.
- Preserve existing parsing behavior for valid non-empty JSONL records.
- Preserve failure behavior for genuinely malformed non-empty JSONL lines.
- Add a focused unit test covering valid JSONL records separated by blank lines.

## Checklist

- [x] I have tested these changes locally.
- [x] I have reviewed the code changes.
- [x] I have updated the documentation (if necessary).
- [x] I have added appropriate unit tests (if applicable).

### What Problem This Solves

A JSONL input file may contain accidental blank or whitespace-only lines between otherwise valid records:

```jsonl
{"text":"a"}

{"text":"b"}
```

Before this change, the blank line was passed directly to `json.loads("")`, which raises `JSONDecodeError`. Because `InputReader._iterate_files()` catches exceptions at the file level, the reader skipped the entire file instead of loading the valid records before and after the blank line.

### Evidence

The affected call chain is:

```text
create_input_reader(...)
  -> JSONLinesFileReader.read_file(...)
    -> text.splitlines()
    -> json.loads(line)
       -> JSONDecodeError on blank line
  -> InputReader._iterate_files(...)
    -> catches exception
    -> logs warning
    -> skips the file
```

This PR only ignores blank or whitespace-only JSONL lines. Malformed non-empty lines still go through `json.loads(line)` and keep the existing error/skip behavior.

### Scope

This change only affects JSONL input handling. It does not change CSV, JSON, text, parquet, MarkItDown, storage behavior, or handling of malformed non-empty JSONL records.

### Validation

- `git diff --check`
- `uv run ruff format --check packages/graphrag-input/graphrag_input/jsonl.py tests/unit/indexing/input/test_jsonl_loader.py`
[...]

PR #2434, opened 2026-07-10
Fix JSONL loader handling of blank/invalid lines

### LINKED ISSUE #2424: JSONL reader skips entire file when input contains blank lines
### Describe the issue

`JSONLinesFileReader.read_file()` parses every line returned by `text.splitlines()` with `json.loads(line)`. If a `.jsonl` file contains an accidental blank or whitespace-only line between valid records, `json.loads("")` raises `JSONDecodeError`.

That exception bubbles up to `InputReader._iterate_files()`, where the reader catches the file-level exception, logs a warning, and skips the entire file. As a result, valid records before and after the blank line are not loaded.

### Steps to reproduce

Create a JSONL input file:

```jsonl
{"text":"a"}

{"text":"b"}
```

Load it through the JSONL input reader.

### Expected behavior

The blank or whitespace-only line is ignored and the two valid JSONL records are loaded.

### Actual behavior

The blank line raises `JSONDecodeError`, and the parent input reader skips the whole file.

### Additional context

This is a small parser robustness issue. It does not change handling for malformed non-empty JSONL lines.

A proposed fix is available in #2423.

natoverse:
## Summary
- Skip blank/whitespace-only lines in JSONL input loader
- Skip malformed JSON rows safely
- Skip non-object JSON rows
- Add regression test and fixture coverage
- Add semversioner patch entry

Fixes #2424


natoverse:
This replaces https://github.com/microsoft/graphrag/pull/2423 to include more handling of edge cases

tests/unit/indexing/input/test_jsonl_loader.py
```diff
@@ -40,3 +40,20 @@ async def test_jsonl_loader_one_file_with_title():
     documents = await reader.read_files()
     assert len(documents) == 3
```
gaudyb, line 41: for this test, is it better to have this fixed value "3" as a default value on top of the test? to understand better why and also to change it in case you want to test something else.

dworthen (approved):

PR #2436, opened 2026-07-11
Fix #2427: preserve node names in stable_lcc so mixed-case entities aren't dropped from communities

### LINKED ISSUE #2427: [Bug]: use_lcc silently drops case-sensitive entity titles from every community

- [X] I have searched the existing issues and this bug is not already filed.
- [X] My model is hosted on OpenAI or Azure. (N/A — reproduces independent of model provider; this is a pure data-pipeline bug, not an LLM output issue)
- [X] I believe this is a legitimate bug, not just a question.

### Describe the bug

When bringing your own graph (BYOG) with `cluster_graph.use_lcc: true` (the default), entities whose `title` contains lowercase characters are silently dropped from every community — with no warning, error, or log line. `create_communities` reports success and writes N communities, but coverage across all communities can be far less than the number of input entities.

Root cause: `graphs/stable_lcc.py`'s `normalize_node_name` does `html.unescape(name).upper().strip()` before computing the largest connected component, and this uppercased form is what gets passed into `hierarchical_leiden` — so the cluster labels that come back out of `cluster_graph()` are uppercase. `index/workflows/create_communities.py`'s `create_communities()` then joins those cluster labels back against `entities.title` via exact case-sensitive string matching (`title_to_entity_id` dict lookup, and `relationships.merge(level_comms, left_on="source", right_on="title")`). Any title that wasn't already all-uppercase fails to match: `entity_map["entity_id"]` comes back `NaN` and is dropped via `.dropna()`, and any community with no matched intra-community edges is dropped entirely by the final `inner` merge against `entity_ids`.

This is presumably fine for GraphRAG's own LLM-extraction pipeline, where entity names are conventionally emitted uppercase already — but it's a silent correctness bug for BYOG users bringing case-sensitive titles (e.g. a codebase's real identifier names, which is a very natural BYOG use case).

### Steps to reproduce

1. Build a minimal BYOG `entities.parquet` / `relationships.parquet` per the [BYOG guide](https://microsoft.github.io/graphrag/index/byog/) with at least one entity `title` containing lowercase letters (e.g. `FooBar`, `BazQux`), and a relationship between two such entities.
2. Configure `workflows: [create_communities, create_community_reports]` with default `cluster_graph.use_lcc: true`.
3. Run `graphrag index`.
4. Inspect `output/communities.parquet` — the mixed-case entities never appear in any community's `entity_ids`, despite being connected in `relationships.parquet`.

### Expected Behavior

Every entity that appears in a connected relationship should be assigned to some community, or the pipeline should surface a warning when input entities can't be matched back after LCC normalization.

### GraphRAG Config Used

```yaml
cluster_graph:
  max_cluster_size: 10
  use_lcc: true  # default - triggers the bug
workflows: [create_communities, create_community_reports, generate_text_embeddings]
```

### Logs and screenshots

N/A — the bug produces no error output; it is a silent data-loss issue only visible by comparing entity counts in `entities.parquet` against total `entity_ids` coverage across `communities.parquet`.

### Additional Information

- GraphRAG Version: 3.1.0
- Operating System: macOS 26.5.1
- Python Version: 3.12.13
- Related Issues: none found

khaledalam:
## Summary
Fixes #2427. With `cluster_graph.use_lcc: true` (the default), Bring-Your-Own-Graph entities whose titles contain lowercase characters were silently dropped from every community. The pipeline reported success but produced incomplete coverage.

## Root cause
`stable_lcc` (`graphs/stable_lcc.py`) uppercased / HTML-unescaped / stripped every node name before computing the largest connected component, and those mutated names became the cluster labels. Downstream, `create_communities.py` matches cluster labels against the original entity titles with an exact, case-sensitive lookup and merges on `source`/`target`. Any mixed-case title (e.g. `FooBar`) never matched its uppercased label (`FOOBAR`), produced `NaN`, and was dropped.

This normalization was a legacy carryover from the old NetworkX implementation. The standard extraction pipeline already emits uppercase titles, so it was a no-op there and only ever corrupted case-sensitive (BYOG) input.

## Fix
`stable_lcc` now preserves node names verbatim and only does its actual job: filter to the largest connected component and stabilize edge ordering/direction.

### DIFF
--- a/.semversioner/next-release/patch-20260711111442208341.json
+++ b/.semversioner/next-release/patch-20260711111442208341.json
@@ -1,0 +1,4 @@
+{
+  "type": "patch",
+  "description": "Fix #2427 - preserve original node names in stable_lcc so case-sensitive BYOG entity titles are not dropped from communities."
+}
--- a/packages/graphrag/graphrag/graphs/stable_lcc.py
+++ b/packages/graphrag/graphrag/graphs/stable_lcc.py
@@ -7,12 +7,14 @@
 regardless of the original row order.  This is achieved by:
 
 1. Filtering to the largest connected component.
-2. Normalizing node names (HTML unescape, uppercase, strip).
-3. Sorting edges so the lesser node is always the source.
-4. Sorting edges alphabetically for deterministic row order.
-"""
+2. Sorting edges so the lesser node is always the source.
+3. Sorting edges alphabetically for deterministic row order.
 
-import html
+Node names are preserved verbatim.  Downstream steps (e.g. create_communities)
+match the resulting cluster labels against the original entity titles with an
+exact, case-sensitive lookup, so any mutation of node names here would silently
+drop entities whose titles are not already normalized (see issue #2427).
+"""
 
 import pandas as pd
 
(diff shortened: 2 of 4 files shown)


linhongyu510:
I checked out this PR and independently ran the focused graph/community test suite:

```text
pytest tests/unit/graphs/test_stable_lcc.py \
       tests/unit/indexing/test_create_communities.py -q
39 passed
```

The implementation keeps edge orientation and row ordering deterministic while preserving the exact node labels used by the downstream case-sensitive entity lookup. I also reviewed the mixed-case regression coverage in both `test_stable_lcc.py` and `test_create_communities.py`; it exercises the failure path described in #2427.

I did not find a correctness regression in the changed scope.

PR #2428, opened 2026-07-09
Add MiniMax M3 model costs

octo-patch:
Reason: add target provider/model to existing provider registry

## Summary
- Add MiniMax M3 cost metadata to the GraphRAG LLM model cost registry.
- Include cache-read pricing, 1M context metadata, and a regression test for the registry entry.

## Test plan
- `git add -A -N && git diff HEAD | grep -E "$SECRET_RE"` (no matches)
- `uv sync --all-packages`
- `PYTHONPATH="/root/octopatch-4/work/repos/microsoft_graphrag/packages/graphrag-llm:/root/octopatch-4/work/repos/microsoft_graphrag/packages/graphrag-cache:/root/octopatch-4/work/repos/microsoft_graphrag/packages/graphrag-common:$PYTHONPATH" uv run pytest tests/integration/language_model/test_factory.py -q`
- `uv run ruff check packages/graphrag-llm/graphrag_llm/model_cost_registry/model_cost_registry.py tests/integration/language_model/test_factory.py`

PR #2438, opened 2026-07-13
Add MiniMax M3 and M2.7 model metadata

octo-patch:
## Description

Add MiniMax-M3 and MiniMax-M2.7 metadata to GraphRAG's model cost registry. The metadata covers context windows, supported modalities, reasoning, prompt caching, and current token pricing, including MiniMax-M3 long-context and priority tiers.

This replaces #2428 with a clean branch based on the latest `main` and complete target coverage.

## Proposed Changes

- Register both model IDs for OpenAI-compatible and Anthropic-compatible configurations.
- Calculate cached, long-context, and priority token costs from response usage.
- Document the global and China base URLs for both supported protocols.
- Add focused regression tests and a patch release note.

## Validation

- `uv run poe check`
- `uv run pytest tests/unit/model_cost_registry/test_model_cost_registry.py tests/integration/language_model/test_factory.py -q` (`12 passed`)
- `uv run poe test_unit` (`310 passed`)
- `uv run poe build`
- `uv run poe build_docs`
- Captured requests through the GraphRAG adapter for both models across all four documented endpoint configurations.

## Checklist

- [x] I have updated the documentation.

### DIFF
--- a/.semversioner/next-release/patch-20260713065517820226.json
+++ b/.semversioner/next-release/patch-20260713065517820226.json
@@ -1,0 +1,4 @@
+{
+  "type": "patch",
+  "description": "Add MiniMax M3 and M2.7 model cost metadata."
+}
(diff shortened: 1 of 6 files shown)


natoverse:
Our cost estimates come for the lookup table in litellm so we don't need to maintain the mapping. This would be best submitted to them, or run separately as a post-process using the token counts emitted in the logs

